PR #8, opened 2019-01-08
Create shaft_element

rodrigomoliveira1:
This file creates the shaft elements


shaft_element
```diff
@@ -0,0 +1,478 @@
+import warnings
+import numpy as np
+import pandas as pd
+import scipy.interpolate as interpolate
+import matplotlib.pyplot as plt
+import matplotlib.patches as mpatches
+from ross.data_io.read_xl import (
```
raphaeltimbo: These imports can be deleted since we have not implemented any data io yet.

shaft_element
```diff
@@ -0,0 +1,478 @@
...
+
+"""The path "ross.data_io.read_x1" on line 7 maybe changes"""
+
+
+__all__ = [
+    "ShaftElement"
+]
+
+
+class Element:
```
raphaeltimbo: Implement `class Element` as `ABC`. 
The docs for ABC can be found [here](https://docs.python.org/3/library/abc.html) and an interesting article [here](https://dbader.org/blog/abstract-base-classes-in-python).

shaft_element
```diff
@@ -0,0 +1,478 @@
...
+        le = L / ne
+
+        elements = [
+            cls(le, si_d, so_d, material, n, shear_effects, rotary_inertia, gyroscopic)
+            for _ in range(ne)
+        ]
+
+        return elements
+
+    @classmethod
```
raphaeltimbo: This `@classmethod` can be removed for now since we are going to implement io later.

rodrigomoliveira1:
i'm already working on it!

rodrigomoliveira1:
i have modified the file "shaft_element" as requested. I replaced the file with the correct extension .py.

shaft_element.py
```diff
@@ -1,478 +1,461 @@
...
+    def summary(self):
+        """A summary for the element.
+        A pandas series with the element properties as variables.
+        """
+        attributes = self.__dict__
+        attributes["type"] = self.__class__.__name__
+        return pd.Series(attributes)
+
+    # These are the abstract classes for mass, damping and stiffness matrices
+    def M(self):
```
raphaeltimbo: The decorator `@abstractmethod` is missing.
The purpose of the decorator is that, when inheriting from `element`, the subclass will have to implement the abstract method. If the method is not implemented a `TypeError` is raised.

shaft_element.py
```diff
@@ -1,478 +1,461 @@
...
+
+    def summary(self):
+        """A summary for the element.
+        A pandas series with the element properties as variables.
+        """
+        attributes = self.__dict__
+        attributes["type"] = self.__class__.__name__
+        return pd.Series(attributes)
+
+    # These are the abstract classes for mass, damping and stiffness matrices
```
raphaeltimbo: After adding the `@abstractmethod` decorator I think you can remove this comment since it will be redundant.

shaft_element.py
```diff
@@ -1,478 +1,461 @@
...
+        Torque.
+    shear_effects : bool
+        Determine if shear effects are taken into account.
+        Default is False.
+    rotary_inertia : bool
+        Determine if rotary_inertia effects are taken into account.
+        Default is False.
+    gyroscopic : bool
+        Determine if gyroscopic effects are taken into account.
+        Default is False.
```
raphaeltimbo: Add `shear_method_calc` as an argument to the docstring.

shaft_element.py
```diff
@@ -1,478 +1,461 @@
...
+        i_d,
+        o_d,
+        material,
+        n=None,
+        axial_force=0,
+        torque=0,
+        shear_effects=True,
+        rotary_inertia=True,
+        gyroscopic=True,
+        shear_method_calc=None,
```
raphaeltimbo: It would be better to leave one of the methods as the default:
`shear_method_calc='cowper'`

shaft_element.py
```diff
@@ -1,478 +1,461 @@
...
+        # diameters
+        self.i_d = float(i_d)
+        self.o_d = float(o_d)
+        self.i_d_l = float(i_d)
+        self.o_d_l = float(o_d)
+        self.i_d_r = float(i_d)
+        self.o_d_r = float(o_d)
+
+        self.material = material
+        self.material_name = material.name
```
raphaeltimbo: Remove these attributes:

```python
self.material_name = material.name
self.E = material.E
self.G_s = material.G_s
self.Poisson = material.Poisson
self.color = material.color
self.rho = material.rho
```
since we are going to access them through the material object.

shaft_element.py
rodrigomoliveira1: Just added the abstract methods.
These changes are shown in this commit [here](https://github.com/rodrigomoliveira1/ross/commit/189ddf53bdb508cd3dc2d61cca62c4182a6e2269)
I will address every commit link in each comment you've made, and the pull request will be done only after the last change.

raphaeltimbo: Great! The commits are already showing in this pull request, so there is no need to make another one.
Just one note on the commit title, it would be better if it was less generic. Instead of 'Update shaft_element.py' something as 'Add abstract methods'.

rodrigomoliveira1: Noted!

shaft_element.py
rodrigomoliveira1: These changes are show in this commit [here](https://github.com/rodrigomoliveira1/ross/commit/3a53437df3e433bd71bbaf316b06eac3dc14e561)

shaft_element.py
rodrigomoliveira1: These changes are show in this commit [here](https://github.com/rodrigomoliveira1/ross/commit/3a53437df3e433bd71bbaf316b06eac3dc14e561)

shaft_element.py
rodrigomoliveira1: These changes are shown in this commit [here](https://github.com/rodrigomoliveira1/ross/commit/100f3bb8f6ba6076a7c77500c59c8cbb82ae218a).

PR #11, opened 2019-01-08
Create materials.py

rodrigomoliveira1:
This file adds different types of materials. in the future, new materials may be added.


raphaeltimbo:
I will close this PR since the materials will be implemented according to #12 .

PR #16, opened 2019-01-15
Implementation of  the Material class

GabrielBachmann:
Here is the implementation of the Material class that can be saved in a JSON format.

### DIFF
--- a/ross/__init__.py
+++ b/ross/__init__.py
@@ -1,0 +1,1 @@
+import ross.materials as materials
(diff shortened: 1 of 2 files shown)


raphaeltimbo:
I will merge this now.
Some comments:
- We will need to add tests to this module;
- Evaluate if we could write `def __init__(self, name, rho=None, E=None, G_s=None):' this way the function signature will appear to the user and facilitate the use;
- Add documentation for the methods.

Issue #19, opened 2019-01-17
Class element is repeated on different modules


raphaeltimbo:
The `class Element` is repeated on different files: shaft_element.py, disk_element.py...
I think we should have that class in a separate element.py file and the other modules should import from it.

rodrigomoliveira1:
Sounds good! i'm doing it right now. Just moved the `class Element` to a new element.py and removed it from the orignials ones.

rodrigomoliveira1:
A new file element.py has been added. The commit is [here](https://github.com/ross-rotordynamics/ross/pull/20/commits/e53c436a86118c414b1ce3ebe3eab64b9bab6945)

rodrigomoliveira1:
So, I just commited changes in 3 files removing the `class element` from them. They're listed below:

- [shaft_element.py](https://github.com/ross-rotordynamics/ross/commit/bc534ec07c4eaba36f156cc145efc80379bf7b0d)

- [disk_element.py](https://github.com/ross-rotordynamics/ross/commit/330745446bc27c518e669333657a7adf4e31107d)

- [bearing_seal_element.py](https://github.com/ross-rotordynamics/ross/commit/5f2a75da7b2ac31f59cf3d76475d3786ac6d4009)

They are all together in a new branch called "element_file" including the element.py file

PR #20, opened 2019-01-17
Add element.py file

rodrigomoliveira1:
This file should replace the main `class Element` from the shaft_element.py, disk_element.py, bearing_seal_element.py files

### DIFF
--- a/ross/bearing_seal_element.py
+++ b/ross/bearing_seal_element.py
@@ -5,33 +5,13 @@ import scipy.interpolate as interpolate
 import matplotlib.pyplot as plt
 import matplotlib.patches as mpatches
 from abc import ABC
+from ross.ross.element.py import *
 
 __all__ = [
     "BearingElement",
     "SealElement",
 ]
-
-class Element(ABC):
-    """Element class."""
-
-    def __init__(self):
-        pass
-
-    def summary(self):
-        """A summary for the element.
-        A pandas series with the element properties as variables.
-        """
-        attributes = self.__dict__
-        attributes["type"] = self.__class__.__name__
-        return pd.Series(attributes)
-
-    @ABC.abstractmethod
-    def C(self):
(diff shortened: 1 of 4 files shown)


ross/bearing_seal_element.py
```diff
@@ -5,33 +5,13 @@
 import matplotlib.pyplot as plt
 import matplotlib.patches as mpatches
 from abc import ABC
+from ross.ross.element.py import *
```
raphaeltimbo: I don't think this import will work. 
Try:
`from element import Element`
In this case is a good practice to be explicit on the imports and not use `import *`

ross/element.py
```diff
@@ -0,0 +1,31 @@
+import pandas as pd
+from abc import ABC
```
raphaeltimbo: The `abstractmethod` should be imported from abc:
```python
from abc import ABC, abstractmethod
```

ross/bearing_seal_element.py
rodrigomoliveira1: Done!

- [shaft_element.py](https://github.com/ross-rotordynamics/ross/commit/14e8f8b2b701d02012c5228baabd522d6897253f)

- [disk_element.py](https://github.com/ross-rotordynamics/ross/commit/690e8ae9f7034da69b4dec034113f323fce429b9)

- [beraing_seal_element.py](https://github.com/ross-rotordynamics/ross/commit/65c71fba65aecd9772d53627faa5bbe5ea807ef5)

ross/element.py
rodrigomoliveira1: Done.

- [element.py](https://github.com/ross-rotordynamics/ross/commit/9bc56eeef8c788cc605ab47ae84fe549eed804e1)

raphaeltimbo:
Closes #19 .

PR #57, opened 2019-03-01
Add static analysis method

rodrigomoliveira1:
Add a method to calculate the static displacement of a rotor due to disks' weight.

### DIFF
--- a/ross/rotor_assembly.py
+++ b/ross/rotor_assembly.py
@@ -217,6 +217,8 @@ class Rotor(object):
         df = df.sort_values(by="n_l")
         df = df.reset_index(drop=True)
 
+        self.df_disks = df_disks
+
         # check consistence for disks and bearings location
         if df.n_l.max() > df[df.type == "ShaftElement"].n_r.max():
             raise ValueError("Trying to set disk or bearing outside shaft")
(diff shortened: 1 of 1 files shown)


ross/rotor_assembly.py
```diff
@@ -1389,6 +1391,28 @@ def load(file_name):
...
+        """
+        # gravity aceleration vector
+        grav = np.zeros((len(self.M()), 1))
+
+        # place gravity effect on disk nodes
+        for i in self.df_disks["n"]:
+            grav[4 * i - 3] = -9.8065
+
+        # calculates x, for [K]*[x] = [M]*[g]
+        disp = la.solve(self.K(0), np.matmul(self.M(), grav))
```
raphaeltimbo: Instead of np.matmul we could use the python matmul operator `@`

```python
disp = la.solve(self.K(0), self.M() @ grav)
```

ross/rotor_assembly.py
```diff
@@ -1389,6 +1391,28 @@ def load(file_name):
...
 
+    def static(self):
+        """This function calculates the static displacement of a rotor
+        due the disks' weight, and returns an array with the nodal displacement
+        The calculation is done from the stiffness and mass matrix of a rotor.
+        """
+        # gravity aceleration vector
+        grav = np.zeros((len(self.M()), 1))
+
+        # place gravity effect on disk nodes
```
raphaeltimbo: I think we should find a way to make this part more readable with something like:
```python
for disk in self.disks:
    grav[4 * disk.node_number - 3] = -9.8065
```
This is just an idea...

PR #63, opened 2019-03-13
Add Bokeh plotting style

rodrigomoliveira1:
Add bokeh plot style to "static" method.
Add deformed and undeformed shaft plots.
Add gravity effect to all elements' nodes (vertical direction - "y").
Correct the loop index while calculating the shaft displacement.
Add bokeh plot style to "from_section" classmethod.

### DIFF
--- a/ross/bearing_seal_element.py
+++ b/ross/bearing_seal_element.py
@@ -209,7 +209,7 @@ class BearingElement(Element):
 
         return C
 
-    def patch(self, ax, position, axis):
+    def patch(self, position, axis):
         """Bearing element patch.
         Patch that will be used to draw the bearing element.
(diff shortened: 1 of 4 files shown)


rodrigomoliveira1:
It is done. I also updated the patches' docstrings and changed the bokeh plot axis' name to bk_ax to avoid misunderstanding the matplotlib axis. The build for each plot is done separately and they are all commented along the code.

rodrigomoliveira1:
Just added a new bokeh plot.
It shows the shaft weight, disk weight and bearings' reaction forces.
This is how it's looking right now, for example:

FBD

Tell me if I need to change something to make it look better.

Issue #65, opened 2019-03-19
Add analytical fluid flow routine for short bearings


flaviorangel:
We decided to create a fluid flow routine that can be calculated analytically before the numerical one.

flaviorangel:
Just an update to this issue's status: although the routine was merged, we still need to make sure it is doing what it is supposed to do. @JuliaMota and professor @juvianna, along with myself, are taking care of it. I'm closing this issue and updating #37 to include what needs to be done related to this one.

PR #66, opened 2019-03-19
Analytical fluid flow, short bearings

flaviorangel:
Add analytical fluid flow routines for short bearings, related to issue #65


ross/fluid_flow/fluid_flow.py
```diff
@@ -0,0 +1,400 @@
...
+            for i in range(self.nz):
+                for j in range(self.ntheta):
+                    self.p_mat[i][j] = ((-3*self.visc*self.omega)/self.difference_between_radius**2)*\
+                                       ((i*self.dz)**2 - (self.length**2)/4)*\
+                                       (self.eccentricity_ratio*np.sin(j*self.dtheta))/\
+                                       (1 + self.eccentricity_ratio*np.sin(j*self.dtheta))**3
+            self.pressure_matrix_available = True
+        return self.p_mat
+
+    def calculate_coefficients(self, plot_cut=False):
```
raphaeltimbo: This function is doing some calculation and making a plot. I think it would be better to separate that and make a different function for the plot functionality.
Additionally, when using a function to create a plot with matplotlib, it is better to make the axes an argument:
```python
def foo(ax=None):
    # if axes is not given get the one from the current figure
    if ax is None:
        ax = plt.gca()
    .
    .
    ax.plot(...)
    .
    .
    return ax
```
For more details about matplotlib objects (figure, axes, axis etc.) refer to https://matplotlib.org/tutorials/introductory/usage.html

ross/fluid_flow/fluid_flow.py
```diff
@@ -0,0 +1,400 @@
...
+        else:
+            self.bearing_seal_type = 2
+        if plot_cut:
+            plt.figure(self.plot_counter)
+            self.plot_counter += 1
+        for i in range(self.nz):
+            zno = i * self.dz
+            self.z[0][i] = zno
+            for j in range(self.ntheta):
+                gama = j * self.dtheta
```
raphaeltimbo: Use `gamma` for consistency.

ross/fluid_flow/fluid_flow.py
```diff
@@ -0,0 +1,400 @@
...
+
+    def internal_radius_function(self, z, gama):
+        """This function calculates the radius of the rotor given the
+        radius and the position z.
+        Parameters
+        ----------
+        z: float
+            Distance along the z-axis.
+        gama: float
+            Gama is the distance in the theta-axis. It should range from 0 to 2*np.pi.
```
raphaeltimbo: Add the following information to the docstring:
```python
"""This function calculates the radius of the rotor given the
...
    Returns
    -------
    radius_external: float
        <info>
    xre: float
        <info>
    yre: float
        <info>
```

ross/fluid_flow/fluid_flow.py
```diff
@@ -0,0 +1,400 @@
...
+        for i in range(0, self.nz):
+            plt.plot(i*self.dz, self.p_mat[i][0], 'bo')
+        plt.title('Pressure along the Z direction (direction of flow); Theta=0')
+        plt.xlabel('Points along Z')
+        plt.ylabel('Pressure')
+        plt.show(block=show_immediately)
+        if show_immediately:
+            plt.close('all')
+
+    def plot_shape(self, theta=0, show_immediately=True):
```
raphaeltimbo: When using a function to create a plot with matplotlib, it is better to make the axes an argument:
```python
def foo(ax=None):
    # if axes is not given get the one from the current figure
    if ax is None:
        ax = plt.gca()
    .
    .
    ax.plot(...)
    .
    .
    return ax
```
For more details about matplotlib objects (figure, axes, axis etc.) refer to https://matplotlib.org/tutorials/introductory/usage.html

ross/fluid_flow/fluid_flow.py
```diff
@@ -0,0 +1,400 @@
...
+        plt.plot(x, y_re, 'r')
+        plt.plot(x, y_ri, 'b')
+        plt.title('Shapes of stator and rotor along Z; Theta='+str(theta))
+        plt.xlabel('Points along Z')
+        plt.ylabel('Radial direction')
+        plt.show(block=show_immediately)
+        if show_immediately:
+            plt.close('all')
+
+    def plot_pressure_theta_cylindrical(self, z=0, show_immediately=True):
```
raphaeltimbo: When using a function to create a plot with matplotlib, it is better to make the axes an argument:
```python
def foo(ax=None):
    # if axes is not given get the one from the current figure
    if ax is None:
        ax = plt.gca()
    .
    .
    ax.plot(...)
    .
    .
    return ax
```
For more details about matplotlib objects (figure, axes, axis etc.) refer to https://matplotlib.org/tutorials/introductory/usage.html

ross/fluid_flow/fluid_flow.py
```diff
@@ -0,0 +1,400 @@
...
+                z_matrix[i][j] = pressure_along_theta[i] - min_pressure + 0.01
+        fig, ax = plt.subplots(subplot_kw=dict(projection='polar'))
+        self.plot_counter += 1
+        ax.contourf(theta_matrix, r_matrix, z_matrix, cmap='coolwarm')
+        plt.title('Pressure along Theta; Z='+str(z))
+        plt.show(block=show_immediately)
+        if show_immediately:
+            plt.close('all')
+
+    def plot_pressure_theta(self, z=0, show_immediately=True):
```
raphaeltimbo: When using a function to create a plot with matplotlib, it is better to make the axes an argument:
```python
def foo(ax=None):
    # if axes is not given get the one from the current figure
    if ax is None:
        ax = plt.gca()
    .
    .
    ax.plot(...)
    .
    .
    return ax
```
For more details about matplotlib objects (figure, axes, axis etc.) refer to https://matplotlib.org/tutorials/introductory/usage.html

ross/fluid_flow/example_of_fluid_flow.py
```diff
@@ -0,0 +1,46 @@
+"""Example of fluid flow
```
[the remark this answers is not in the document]
flaviorangel: I agree. I'm removing this file.

ross/fluid_flow/fluid_flow.py
```diff
@@ -0,0 +1,400 @@
...
+        of the discrete pressure (central differences in the second
+        derivatives). It is executed when the class is instantiated.
+        Parameters
+        ----------
+        plot_cut: bool
+            If True, plots a cut at z=0 and show, so the user can check the
+            internal and external radius as well as eccentricity.
+        """
+        if self.length/self.radius_stator <= 1/8:
+            self.bearing_seal_type = 0
```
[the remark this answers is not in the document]
flaviorangel: Ok. I'm changing to 'short_size', 'long_size', and 'medium_size'. I did not use 'short bearing' because it can be either a bearing or a seal. Is there a name that includes both?

ross/fluid_flow/fluid_flow.py
flaviorangel: Did you mean checking if gama is between 0 and 2*pi? If so, ntheta is provided by the user when instantiating, dtheta is calculated automatically based on ntheta, therefore gama can't be wrong.

ross/fluid_flow/fluid_flow.py
flaviorangel: Ok. I'm working on it.

flaviorangel: I'm working on creating a new plot function to separate it from this function.
As for these changes in ploting functions, we will soon change from matplotlib to bokeh, so I think we can let them like this for now. What do you think?

ross/fluid_flow/fluid_flow.py
flaviorangel: Done.

ross/fluid_flow/fluid_flow.py
flaviorangel: We will soon change from matplotlib to bokeh, so I think we can let them like this for now. What do you think?

ross/fluid_flow/fluid_flow.py
flaviorangel: We will soon change from matplotlib to bokeh, so I think we can let them like this for now. What do you think?

ross/fluid_flow/fluid_flow.py
flaviorangel: We will soon change from matplotlib to bokeh, so I think we can let them like this for now. What do you think?

ross/fluid_flow/fluid_flow.py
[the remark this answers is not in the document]
JuliaMota: For now our code can only be used for bearings. In future we can include the seal in another code. I agree we can change the name to "short_bearing" to be more descriptive.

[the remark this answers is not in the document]
flaviorangel: Ok. I changed the names back to bearing and removed any mention to "seal" from the code.

raphaeltimbo:
I am going to merge this now. 
Regarding the matplotlib plots, I think we need to debate if we will use only bokeh or if we are going to provide both options. My only doubt is if with bokeh we can generate plot files that are ok for publication (svg or pdf?).

PR #78, opened 2019-04-02
Remove plot when creating rotor

raphaeltimbo:
The rotor geometry was plotted when the object was instantiated. The
purpose of this was to show the user the rotor geometry so that the user
could check if everything was correct before eigenvalues, eigenvectors
etc. were calculated.
The problem with this approach is that the rotor will be automatically
plotted in situations where this is not wanted by the user, such as when
running a script where several rotors will be instantiated or when we
are running the tests locally, in this case, several windows with plots
will appear and the tests will take a long time.
With this PR the plot of the rotor after instantiation will be optional
to the user. A message is provided to let the user know that the
plot_rotor() method can be used.
Additionally, the message "calculate eigenvalues and frequencies", which
is redundant, has been modified to "calculate rotor state".

### DIFF
--- a/ross/rotor_assembly.py
+++ b/ross/rotor_assembly.py
@@ -279,8 +279,8 @@ class Rotor(object):
 
         #  diameter at node position
 
-        self.plot_rotor()
-        print("To calculate eigenvalues and frequencies, use the method run().")
+        print("To check the rotor geometry, use the method plot_rotor()\n"
+              "To calculate the rotor state, use the method run()")
 
     def __eq__(self, other):
         if self.elements == other.elements and self.parameters == other.parameters:


GabrielBachmann:
What I had in mind to solve this issue, is to add a new parameter(test_mode=False) to the Rotor class and put an if to properly plot the rotor(if test_mode==False),  what do you think?

raphaeltimbo:
> What I had in mind to solve this issue, is to add a new parameter(test_mode=False) to the Rotor class and put an if to properly plot the rotor(if test_mode==False), what do you think?

That could be an option, but I am worried if this will be intuitive for the user in the case where they are running a script that creates multiple rotors.

I am saying that because I believe that one of the main purpose of our project is to make the library to be called by other projects/scripts and in this case the graphical interface would be secondary, and I consider automatically showing the plot a kind of graphical interface.

@rodrigomoliveira1 @thiagoritto what do you guys think about this issue?

rodrigomoliveira1:
Well, indeed we need an option to avoid this kind of situation. For example, when we run "convergence" classmethod, it plots every rotor built at each iteration, and this might be annoying for the user, moreover, it consumes more RAM while plotting.
I think @raphaeltimbo's option is better because it is easier for the user to understand the command required to plot the rotor model. On the other hand, adding a parameter like @GabrielBachmann said allows the user to check every model build at each iteration, if the user needs to check something in real time. The fisrt option won't allow you retrieve everything, only the last one done, I guess. But that's why we are adding tests files. So, I am with @raphaeltimbo.

PR #85, opened 2019-04-10
Set parameters (static method) as object's attribute

rodrigomoliveira1:
In the static mehtod, set 3 parameters as object's attribute. The main reason for doing it is that we can return and reffer to them later if needed. (pytest precisely).
disp_y - displacement in y direction (array);
Vx - shearing force (array);
Bm - Bending moment (array)


rodrigomoliveira1:
The difference between the results is due to the incorrect Kyy value for the bearing. The results calculated by hand were for a Kyy = 1e6 N/m and in the program (rotor3) was Kyy = 0.8e6 N/m.

ross/bearing_seal_element.py
```diff
@@ -109,6 +111,35 @@ class BearingElement(Element):
...
+    >>> # following stiffness and damping coefficients and speed range from
+    >>> # 0 to 200 rad/s
+    >>> Kxx = 1e6
+    >>> Kyy = 0.8e6
+    >>> Cxx = 2e2
+    >>> Cyy = 1.5e2
+    >>> W = np.linspace(0,200,11)
+    >>> bearing0 = BearingElement(n=0, kxx=Kxx, kyy=Kyy, cxx=Cxx, cyy=Cyy, w=W)
+    >>> bearing0.K(W)
+    array([[[1000000., 1000000., 1000000., 1000000., 1000000., 1000000.,
```
raphaeltimbo: I recommend using `# doctest: +ELLIPSIS` on the docstring examples where we have a large size array as output.
Another option is to slice the output.
This way we avoid having a large array cluttering the docstring.

ross/bearing_seal_element.py
```diff
@@ -109,6 +111,35 @@ class BearingElement(Element):
...
     --------
+    >>> # A bearing element located in the first rotor node, with these
+    >>> # following stiffness and damping coefficients and speed range from
+    >>> # 0 to 200 rad/s
+    >>> Kxx = 1e6
+    >>> Kyy = 0.8e6
+    >>> Cxx = 2e2
+    >>> Cyy = 1.5e2
+    >>> W = np.linspace(0,200,11)
+    >>> bearing0 = BearingElement(n=0, kxx=Kxx, kyy=Kyy, cxx=Cxx, cyy=Cyy, w=W)
```
raphaeltimbo: I think we should standardize the examples with the 
```python
import ross as rs
```
So here we would have 
```python
bearing0 = rs.BearingElement(…)
```

raphaeltimbo: I think we should standardize the examples with the 
```python
import ross as rs
```
So here we would have 
```python
bearing0 = rs.BearingElement(…)
```

raphaeltimbo: I think we should standardize the examples with the 
```python
import ross as rs
```
So here we would have 
```python
bearing0 = rs.BearingElement(…)
```

rodrigomoliveira1:
Now the new tests for static analysis are working just fine.

rodrigomoliveira1:
The PR #109 is working now. The commits done here now follows that new PR. I'm closing this one.

PR #92, opened 2019-04-21
Fixing the save_bearing_seal_element() method

GabrielBachmann:
There was a mistake while saving the coefficients to toml format.

### DIFF
--- a/ross/bearing_seal_element.py
+++ b/ross/bearing_seal_element.py
@@ -186,22 +186,22 @@ class BearingElement(Element):
             
     def save_bearing_seal_element(self, file_name):
         data = Rotor.load_data(file_name)
-        if type(element.w) == np.ndarray:
+        if type(self.w) == np.ndarray:
             try:
-                element.w[0]
+                self.w[0]
                 w = list(element.w)
             except IndexError:
                 w = None
         data[file_name[:-6]][str(element.n)] = {
             "n": element.n,
-            "kxx": element.kxx.coefficient,
-            "cxx": element.cxx.coefficient,
-            "kyy": element.kyy.coefficient,
-            "kxy": element.kxy.coefficient,
-            "kyx": element.kyx.coefficient,
-            "cyy": element.cyy.coefficient,
-            "cxy": element.cxy.coefficient,
-            "cyx": element.cyx.coefficient,
+            "kxx": self.kxx.coefficient,
+            "cxx": self.cxx.coefficient,
+            "kyy": self.kyy.coefficient,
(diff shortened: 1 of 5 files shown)


raphaeltimbo:
This is good for me.
Some suggestions:
- Move methods such as `save_bearing_seal_element()` to their specific class (`class BearingElement` in this case). This way the methods will be standard and not static methods. Also, by doing this I think we could add the `save()` method to the `class Element` ABC, forcing the implementation of the method in new classes.
- In some places, the `Rotor.` class is being called. Maybe we could use `self.`?
- Docstring says it will save the rotor to a binary file which gives the impression that the file is not human readable. I think we could change this to 'toml file'.

GabrielBachmann:
@raphaeltimbo All the changes requested are now implemented in the branch. The code was basically rewritten in the last commit, now the Element class have two new methods, which are save() and load() (static method). Besides that, I've moved the auxiliary static methods load_data() and dump_data() to the Element class, as I was having import issues maintaining them as Rotor methods. Where do you think we should put it?

GabrielBachmann:
The [AppVeyor build](https://ci.appveyor.com/project/raphaeltimbo/ross/builds/24159076/job/e1b9d0sapafyxjdj) has failed, but only in this python version with the following log report:  
`Uploading artifacts...        
`  
`Error uploading artifact the storage: The remote name could not be resolved: 'appveyorcidatav2.blob.core.windows.net'`

I think the problem is not linked with issues in the code itself.

raphaeltimbo:
> @raphaeltimbo All the changes requested are now implemented in the branch. The code was basically rewritten in the last commit, now the Element class have two new methods, which are save() and load() (static method). Besides that, I've moved the auxiliary static methods load_data() and dump_data() to the Element class, as I was having import issues maintaining them as Rotor methods. Where do you think we should put these methods?

I think the `load_data()` and `dump_data()` methods could be in the `Element` class or in a separate utils module. For the `save()` and `load()` methods, I think they should be abstractmethods in the `Element` class.

raphaeltimbo:
> The [AppVeyor build](https://ci.appveyor.com/project/raphaeltimbo/ross/builds/24159076/job/e1b9d0sapafyxjdj) has failed, but only in python version 3.7 with the following log report:
> `Uploading artifacts... `
> `Error uploading artifact the storage: The remote name could not be resolved: 'appveyorcidatav2.blob.core.windows.net'`
> 
> I think the problem is not linked with issues in the code itself.

I have rebuilt this commit and it is passing on appveyor. I think it was just a problem with appveyor itself, since all the tests are passing.

PR #102, opened 2019-05-03
Add Sommerfeld number function

### LINKED ISSUE #101: Include Sommerfeld number in the fluid flow class
According to Friswell et al. in Dynamics of Rotating Machines (2010, page 178), if the magnitude of the load is known, then the bearing eccentricity may be obtained by solving a quartic equation including the Sommerfeld number. The program must be able to calculate the Sommerfeld number and give the eccentricity of the bearing.

flaviorangel:
Add a function to calculate the Sommerfeld number for a given load. Also add two more functions: one to calculate eccentricity for a given load using the Sommerfeld number, and one to calculate the load using the eccentricity given when the class is instantiated.
Close #101


ross/fluid_flow/fluid_flow.py
```diff
@@ -266,6 +269,41 @@ def external_radius_function(self, gama):
 
         return radius_external, xre, yre
 
+    def get_rotor_load(self):
+        """Returns the load applied to the rotor.
```
raphaeltimbo: I think we should use the [numpydoc docstring standard](https://numpydoc.readthedocs.io/en/latest/format.html#sections).

PR #110, opened 2019-05-10
Add interpolate to rotor_assembly

raphaeltimbo:
The function interp1d is used but the module scipy.interpolate was not
being imported.


rodrigomoliveira1:
I have corrected this problem in PR #109  too.

raphaeltimbo:
> I have corrected this problem in PR #109 too.

I have missed that! Will close this PR here.

PR #112, opened 2019-05-13
Added interpolate to rotor_assembly.py

atbrandao:
The function interp1d was used but the interpolate function was not being
imported.


rodrigomoliveira1:
I have already corrected this problem in PR #109 .

Issue #114, opened 2019-05-16
Improve plot_rotor


atbrandao:
1. The node number tags are evenly spaced along the rotor plot. They should be placed on top of each respective node, and the nodes are usually not evenly spaced.

2. The X and Y coordinates shoul be on the same scale on the plot, so the user can have an idea of the rotor actual proportions.

3. Depending on the number of nodes, the bearing representation becomes disproportionally large. The bearing representation on plot_rotor should have its axial length based on the length of adjacent elements.

4. Indicate the rotor sections that might be hollow (i_d > 0). The rotor plot currently shows the rotor to be always solid.

rodrigomoliveira1:
Pull Request #117 have been merged.

PR #117, opened 2019-05-17
Improving plot_rotor

rodrigomoliveira1:
There was an error that shaft elements were plotting with 2 times their respective radius.
I enlarged the plot window size and distributed the axis size so that the rotor is presented with the correct proportions in relation to its dimensions.

Also, adds the slenderness ratio parameter (equation: G * A * L ** 2 / E * I ) to each shaft element. It calculates the ratio using some global distance measures, rather than basing it upon individual element dimensions. The user is returned an warning message if it's lower than a certain value, which could affect the convergence analysis.

### DIFF
--- a/ross/bearing_seal_element.py
+++ b/ross/bearing_seal_element.py
@@ -289,7 +289,7 @@ class BearingElement(Element):
 
         return C
 
-    def patch(self, position, ax, bk_ax):
+    def patch(self, position, length, ax, bk_ax):
         """Bearing element patch.
         Patch that will be used to draw the bearing element.
         Parameters
@@ -300,6 +300,8 @@ class BearingElement(Element):
             Axes in which the plot will be drawn.
         position : tuple
             Position (z, y) in which the patch will be drawn.
+        length : float
+            minimum length of shaft elements
         Returns
         -------
         ax : matplotlib axes
@@ -308,7 +310,7 @@ class BearingElement(Element):
             Returns the axes object with the plot.
         """
         zpos, ypos = position
-
+        le = length
         h = -0.5 * ypos  # height
 
         #  node (x pos), outer diam. (y pos)
@@ -323,10 +325,10 @@ class BearingElement(Element):
         )
 
         # bokeh plot - upper bearing visual representarion
-        bk_ax.quad(top=-ypos,
-                   bottom=-2 * ypos,
-                   left=zpos - ypos,
-                   right=zpos + ypos,
+        bk_ax.quad(top=-ypos+le/3,
+                   bottom=-ypos,
+                   left=zpos-le/6,
+                   right=zpos+le/6,
                    line_color=bokeh_colors[0],
                    line_width=1,
                    fill_alpha=1,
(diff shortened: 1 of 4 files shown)


rodrigomoliveira1:
Following issue #114, for now, these commits should correct only the issue items 2, 3 and 4.

ross/rotor_assembly.py
```diff
@@ -1208,6 +1208,18 @@ def plot_rotor(self, nodes=1, ax=None, bk_ax=None):
...
         """
+
+        # check slenderness ratio of beam elements
+        SR = np.array([])
+        for shaft in self.shaft_elements:
+            if shaft.sld_ratio < 30:
+                SR = np.append(SR, shaft.n)
+        if len(SR) != 0:
+            print(
+                "Warning: The beam elements " + str(SR) + " have slenderness\n"
```
raphaeltimbo: I think that here it would be better to use warning instead of a print statement:

```python
from warning import warn

...
if len(sr) != 0:
    warn(...)

File: ross/bearing_seal_element.py [hunk 2]
```diff
@@ -300,6 +300,8 @@ def patch(self, position, ax, bk_ax):
             Axes in which the plot will be drawn.
         position : tuple
             Position (z, y) in which the patch will be drawn.
+        length : float
+            minimum length of shaft elements
         Returns
```
raphaeltimbo: It looks like the axes are not being returned, so I think we should delete that from the docstring.

ross/disk_element.py
```diff
@@ -232,16 +235,16 @@ def patch(self, position, ax, bk_ax):
             color=bokeh_colors[9]
         )
         bk_ax.circle(
-            x=zpos,
-            y=ypos + D,
-            radius=0.02,
+            x=bk_disk_points_u[0][0],
```
raphaeltimbo: The indices here (`[0][0]`) make the code hard to read. Maybe this could be better with a dictionary? See if it is possible to improve it.

ross/rotor_assembly.py
```diff
@@ -1231,13 +1247,15 @@ def plot_rotor(self, nodes=1, ax=None, bk_ax=None):
...
-            y_range=[-6 * max_diameter, 6 * max_diameter],
+            tools="pan, wheel_zoom, reset, save",
+            width=1800,
+            height=900,
+            x_range=[-0.1 * shaft_end, 1.1 * shaft_end],
+            y_range=[-0.3 * shaft_end, 0.3 * shaft_end],
             title="Rotor model",
-            x_axis_label="Axial location (m)",
-            y_axis_label="Shaft radius (m)",
+            x_axis_label="Axial location",
```
raphaeltimbo: Shouldn't we keep the units for clarity?

ross/shaft_element.py
```diff
@@ -142,6 +142,10 @@ def __init__(
         self.Ie = np.pi * (o_d ** 4 - i_d ** 4) / 64
         phi = 0
 
+        # Slenderness ratio of beam elements (G*A*L^2) / (E*I)
+        sld = (self.material.G_s * self.A * self.L ** 2) / (self.material.E * self.Ie)
+        self.sld_ratio = sld
```
raphaeltimbo: I think `self.slenderness_ratio = sld` will be more clear to the user.

rodrigomoliveira1: Talking about slenderness ratio (SR), I had an idea to better warn the user if some shaft element has it small enough to might be a problem: plotting these elements in a different color, like a "visual warning". While the elements that are OK are plotted in gray, the elements with SR smaller than 30 are plotted in yellow. Take a look in these plots here: 

Here, all the elements have their SR above 30.
rotor1

and here, some elements have their SR below 30.
rotor2

What do you think about it?

PR #125, opened 2019-05-27
Numerical fluid flow test

JuliaMota:
I have included the numerical pressure calculations for the bearing. This output data can be compared with the analytical pressure


JuliaMota:
The error is in test_fluid_flow, but I'm not sure how to fix it. @flaviorangel , can you help me with this?

flaviorangel:
I deleted the 'testando.py' file because every test should be in the 'tests' folder.

PR #150, opened 2019-06-06
Add release notes to version 0.2.0 - alpha release

raphaeltimbo:
Add release notes based on the pull requests that have the following
labels: 'enhancement', 'api changes' or 'bug'. I have used pygithub to
access all these pull requests and automatically generate an rst file.
I believe we could use this in future releases to automate the process.

I have added everyone in the project as a reviewer to this PR so we can 
check if there is anything missing on the release notes.


rodrigomoliveira1:
What if we include PR #103 as an enhancement? Pointing out the development of an user's guide.

raphaeltimbo:
> What if we include PR #103 as an enhancement? Pointing out the development of an user's guide.

I'll do that!

PR #151, opened 2019-06-06
Add import to Axes3d

raphaeltimbo:
The import has been added to the results module because the mode shape
plot needs this to work. Although it looks like the import is not being
used (this is highlighted by linters), matplotlib, for some strange
reason, actually needs this to use the projection='3d'.
A doctest has been added to the mode shape plot for the tests to fail
if this import is removed.

### DIFF
--- a/ross/results.py
+++ b/ross/results.py
@@ -4,6 +4,7 @@ import bokeh.palettes as bp
 import matplotlib as mpl
 import matplotlib.pyplot as plt
 import numpy as np
+from mpl_toolkits.mplot3d import Axes3D
 from bokeh.layouts import gridplot
 from bokeh.models import ColumnDataSource, ColorBar
 from bokeh.plotting import figure, output_file, show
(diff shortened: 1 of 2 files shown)


ross/results.py
```diff
@@ -637,6 +638,30 @@ def plot(self, dof, ax0=None, ax1=None, **kwargs):
 
 class ModeShapeResults(Results):
     def plot(self, mode=None, evec=None, fig=None, ax=None):
```
GabrielBachmann: When mode value is set to default, the code is returning an Error, as you can see in:

image
I think we can remove this variable default value or change it to 0, what do you think?

GabrielBachmann: This error was obtained by running the following code:
```
import ross as rs 
mode = rs.rotor_example().run_mode_shapes()
```

raphaeltimbo: I have changed mode to be a required positional argument.
If you try: 
```python
import ross as rs 
mode = rs.rotor_example().run_mode_shapes()
mode.plot()
```
it will raise with an error message that the user can understand:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-4-5fd538f603a5> in <module>
----> 1 mode.plot()

TypeError: plot() missing 1 required positional argument: 'mode'
```

PR #153, opened 2019-06-06
Update version to 0.1.0

flaviorangel:
Update init file to version 0.1.0, which is already in PyPI

### DIFF
--- a/ross/__init__.py
+++ b/ross/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.0.2"
+__version__ = "0.1.0"
 
 from .bearing_seal_element import *
 from .disk_element import *


flaviorangel:
@raphaeltimbo I saw a PR mentioning version 0.2.0 and it got me thinking. Shouldn't it be 0.1.1? Also, how do we know in which version we are? We might need new milestones.

raphaeltimbo:
> @raphaeltimbo I saw a PR mentioning version 0.2.0 and it got me thinking. Shouldn't it be 0.1.1? Also, how do we know in which version we are? We might need new milestones.

You are right. I will edit the release notes and add a new milestone for version 0.2.0.

PR #154, opened 2019-06-07
Change plots so that they will not automatically plot

### LINKED ISSUE #140: fluid_flow code automatically showing plots
When running the fluid_flow code, the plots are showing automatically.
We should try only to return the objects and let the users decide if they will plot the results.
For matplotlib, one common way of coding this is the following:
```python
def plot_func(args, ax=None):
    if ax is None:
        # get current axes, if None a new axes is created.
        ax = plt.gca() 
    ...
    ax.plot(...)
    ...
    return ax
```
For bokeh we can return the bokeh figure.

flaviorangel:
Also, among other small changes, plots now distinguish the entry matrix, if numerical or analytical. If both were calculated, they plot both so that the user can compare them.
Close #140

### DIFF
--- a/ross/fluid_flow/fluid_flow.py
+++ b/ross/fluid_flow/fluid_flow.py
@@ -227,8 +227,6 @@ class PressureMatrix:
         self.P = np.zeros([self.ntotal, 1])
         self.p_mat_numerical = np.zeros([self.nz, self.ntheta])
         self.calculate_coefficients()
-        self.mounting_matrix()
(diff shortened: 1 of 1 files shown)


flaviorangel:
The error is within a doctest. The example code is printing results that are different here than in my computer. Probably because of the way the pressure matrix is being computed in each machine. I will erase this line of test in order to approve this PR, but we will need a better example to deal with this.

PR #158, opened 2019-06-10
Add new fluid flow example and fix bearing type definition

flaviorangel:


flaviorangel:
A few commits are duplicated from another PR that was already merged. This is probably because I did not follow the step-by-step when creating a new branch. I started from a folder that was holding another branch. I'll redo this PR then.

PR #178, opened 2019-06-18
Improvements for plotting results

rodrigomoliveira1:
This PR make some modifications in results.py
- Add show(camp) to show campbell bokeh plotting
- Add command to allign ColorBar axis for better looking
- Add figure dimensions for campbell bokeh plotting
- increases the font size of all axis labels 

Some bokeh figures have the attribute `sizing_mode="stretch_both`.  This command line is used so that the plotting of the graphics is always the maximum size of the browse. However, this creates a bug in jupyter notebook, and the graphs does not appears. So it should be replaced  by `width` and `height` command lines

- adds an interpolation function to highlight to the user the intersection between the critical frequency curves and the speed line


rodrigomoliveira1:
My bad @raphaeltimbo. I have thought it was just missing. This is a good idea indeed. So, I'm going to remove it. In fact, I still have to do others changes besides it before merging this PR. I added an interpolation function to check lines intersections, but it doesn't work for multiples harmonics yet.

rodrigomoliveira1:
This is the campbell diagram with interpolation function to find the intersection between the the plots:

Capturar

The red squares mark the intersection, so it should help the user to find it, instead of trying to estimate it visually.

PR #182, opened 2019-06-26
Implementing tests for H_kappa

rodrigomoliveira1:
I'm starting to add some tests for functions that are called when calculating campbell's diagram.
I will let these boxes here to check it for awhile.

Tests implemented:
- [x]  H_kappa
- [x] kappa
- [x] kappa_mode
- [x] whirl_values
- [x] whirl_directions


rodrigomoliveira1:
I implemented some tests for kappa_mode, but I don't know if it's good enough. The values for kappa are always too close to +1 or -1. The difference is something about 1e-10. So, checking the values themselves may be not useful. Evaluating the signals is an option, however the tests for whirl_directions and whirl_values do this job. What you think?

rodrigomoliveira1:
Tests for .H_kappa() are resulting in different values each time it's run. Although it seems worrying, the values are very small (close to 0), as we can see in the following image.

This is the log error from Travis
Capturar

PR #199, opened 2019-07-11
Add model sheet instantiation for shaft element

### LINKED ISSUE #196: Change shaft .from_table() implementation
I have added some tests to the `ShaftElement.from_table()` method (see 895d57927d0775243712f27b6fca7aeba0636b42). Currently, they are marked with `@pytest.mark.skip()` to avoid failures on our CI.
The `.from_table()` method needs to be modified to consider the following:
 - User can use a table with SI or US units. When calling the method, the user should not have to specify which units he is using, this can be extracted from the spreadsheet (cells with 'meters' for example show that we are using SI);
- The method should have a default argument `sheet="Model"`, that points to the sheet inside the spreadsheet that has the model information;

@flaviorangel can you take a look at this? Tables used for the tests are at `/ross/tests/data/`.
After modifying the method we should remove the `@pytest.mark.skip()` from the test.

flaviorangel:
The user can now instantiate shaft elements using a model sheet.
Close #196


raphaeltimbo:
Great! We only need to change the test and replace `Path.cwd()` with another path, since `.cwd()` could be different depending on how we are executing pytest. I think we could use the ross path as a reference, or maybe the test file path.

PR #200, opened 2019-07-14
Update bearing_seal_element.py and UML diagram

GabrielBachmann:
Substitute pytest.approx() to np.allclose() on the bearing __eq___ method.


GabrielBachmann:
Here is the diagram
UML_Diagram_alpha_version (1)

Issue #195, opened 2019-07-09
Create new file - report_API


rodrigomoliveira1:
The objective here is to create a file which receives a rotor as input and run all the required analysis from the Standards. 

The initial ideas for inputing a rotor were:
- input a single rotor for single Standard analysis.
- load previous rotors from toml to run the analysis.

We can propose methods in this thread. I think we could start using these 2 main methods: 
- stability_lvl1: this function will list all the initial steps the Standard requires
- stability_lvl2: this function will list all the following steps the Standard requires, if the rotor fails to meet the requirements of stability_lvl1.

I will open a new branch to upload this file, so we can work together on it.

raphaeltimbo:
I will try to put here the methods and an initial docstring describing the expected behavior:

```python
class Report:
    def __init__(self, rotor):
        """Report according to standard analysis.

        Parameters
        ----------
        rotor : ross.Rotor

        """
        pass

    def unbalance_response(self, clearances):
        """Evaluates the unbalance response for the rotor.

        This analysis takes the critical speeds of interest, calculates the
        position and weight of the required unbalance and performs the analysis
        including:
         - Check if vibration at MCS is below the limit with the applied weight;
         - Check if the clearances are ok if the vibration deteriorate to the
         limit level;

        Parameters
        ----------
        clearances : dict
            Dict mapping between node and its clearance in meters.
            (e.g. clearances = dict(3=0.001, 5=0.002...)
        """

    def stability_level_1(self):
        """Stability analysis level 1.

        This analysis consider a anticipated cross coupling QA based on
        conditions at the normal operating point and the cross-coupling
        required to produce a zero log decrement, Q0.

        Components such as seals and impellers are not considered in this
        analysis.

        Parameters
        ----------
        (Check what we need to calculate the applied cross coupling and list
         them as parameters)
        """
        pass

    def stability_level_1(self):
        """Stability analysis level 2.

        For the level 2 stability analysis additional sources that contribute
        to the rotor stability shall be considered such as:
        a)  labyrinth seals;
        b)  damper seals;
        c)  impeller/blade flow aerodynamic effects;
        d)  internal friction.

        Parameters
        ----------
        (Check what we need to calculate the applied cross coupling and list
        them as parameters)
        """
```

rodrigomoliveira1:
Nice, @raphaeltimbo!
Right now I'm working on an way to plot this graph below, when performing a frequency/unbalance response analysis:

[...]

PR #206, opened 2019-07-15
Add API_report

rodrigomoliveira1:
Work in Progress

This PR adds some plotting styles follwing the API reference:
- Separation Margin
- Amplification Factor 
- Table of API parameters

It follows the Issue #195


raphaeltimbo:
I have added the [wip app](https://github.com/marketplace/wip) to our repository. 

By default, the WIP app sets a pull request status to pending if it finds one of the following terms in its title

- wip
- work in progress
- :construction:

This will block merging.

ross/api_report
```diff
@@ -0,0 +1,538 @@
...
+        Span,
+        Label,
+)
+
+# set bokeh palette of colors
+bokeh_colors = bp.RdGy[11]
+
+
+class Report:
+    def __init__(self, rotor, safety_factor=1.0):
```
raphaeltimbo: What is the use of `safety_factor`? Do we really need this? If so, we should explain in the docstrings.

ross/api_report
```diff
@@ -0,0 +1,538 @@
...
+        [...]
+        """
+
+        N = 60 * maxspeed / (2 * np.pi)
+
+        # get reaction forces on bearings
+        Fb = self.rotor.run_static().force_data['Bearings Reaction Forces']
+        Fb = np.array(Fb) / 9.8065
+
+        # adicionar caso se não houver discos - eixo em balanço
```
raphaeltimbo: Remove the comment.

ross/api_report
```diff
@@ -0,0 +1,538 @@
...
+        """
+        self.rotor = rotor
+        self.safety_factor = safety_factor
+
+    @classmethod
+    def from_saved_rotors(cls, path):
+        rotor = rs.Rotor.load(path)
+        return cls(rotor)
+
+    def static_forces(self, maxspeed, mode):
```
raphaeltimbo: At the end, the primary purpose of this function is to calculate the unbalance forces, so I think we should change the name of the function to something like `.calculate_unbalance_forces()`. Another option is that we split this function in two, the first would be `.static_forces()` and it would return the `Fb` array, and the second would be calculate_unbalance_forces()` and it would use the `Fb` array as argument.

ross/api_report
```diff
@@ -0,0 +1,538 @@
...
+
+        Example
+        -------
+        >>> rotor = rotor_example()
+        >>> report = Report(rotor=rotor)
+        >>> report.mode_shpae(mode=0) # doctest: +ELLIPSIS
+        [...]
+        """
+        modes = (self.rotor.evectors[: self.rotor.ndof]),
+        evec0 = modes[0][:, mode]
```
raphaeltimbo: This code repeats what we have in `ModeShapeResults`. 
Instead of repeating all the code here, we should change `ModeShapeResults` so that the information we need is available for us to use here.

ross/api_report
rodrigomoliveira1: I thought about including this only for user's customization, but the Standards do not use safety factor for anything. I'm removing it.

ross/api_report
rodrigomoliveira1: I repeated the code because I wanted to avoid making modifications to the main program. But following your suggestions, I change what ModeShapesResults returns.

ross/api_report
rodrigomoliveira1: Removed

ross/api_report
rodrigomoliveira1: I did two separated methods:
- one to get the reaction forces
- other to calculate the unbalance forces

ross/api_report
rodrigomoliveira1: I repeated the code because I wanted to avoid making modifications to the main program. But following your suggestions, I change what ModeShapesResults returns.

PR #208, opened 2019-07-16
Fix xl file bug

### LINKED ISSUE #207: Error when trying to load bearing element from xl file
I tried to import an xl file but I was not able to. 
I have added this file to our tests/data folder (see https://github.com/ross-rotordynamics/ross/commit/8eedac2e4f505a13eac5abe0f11599d76c7b5882).
This file should be used to test the .from_table() method implemented
for bearings and seals. Tests should be similar to those that we have
for importing shaft elements from an xl file.

flaviorangel:
The from_table method was too strict. I added a few routines so that the user can vary the format of the table a bit more. I also added a simple test.
Close #207

### DIFF
--- a/ross/bearing_seal_element.py
+++ b/ross/bearing_seal_element.py
@@ -437,17 +437,22 @@ class BearingElement(Element):
         A bearing object.
         """
         try:
-            df = pd.read_excel(file)
+            df = pd.read_excel(file, header=None)
         except FileNotFoundError:
             sys.exit(file + " not found.")
         except xlrd.biffh.XLRDError:
-            df = pd.read_csv(file)
+            df = pd.read_csv(file, header=None)
         header_index = -1
+        header_found = False
         for index, row in df.iterrows():
             for i in range(0, row.size):
                 if isinstance(row[i], str):
-                    header_index = index
-                    break
+                    if row[i].lower() == 'kxx':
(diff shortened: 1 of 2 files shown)


ross/bearing_seal_element.py
```diff
@@ -442,37 +442,62 @@ def from_table(cls, n, file):
             sys.exit(file + " not found.")
         except xlrd.biffh.XLRDError:
             df = pd.read_csv(file)
+        header_index = -1
+        for index, row in df.iterrows():
+            for i in range(0, row.size):
+                if isinstance(row[i], str):
```
raphaeltimbo: To check the header, I think that you should not only check if you have `str`, but also if that string is equal to something (maybe == 'speed'?).
I am saying this because we could have this xl file with some comments on the cell above the actual table.

flaviorangel: You're right.

flaviorangel: In your experience, are these tables of damping and stiffness matrices usually too large? With the current changes, the method searches the entire table for a 'kxx' cell. If these tables are too large, this could get inefficient.

raphaeltimbo: Usually, they are not too large. I think we won't have problems with this being inefficient.

PR #212, opened 2019-07-23
Add from table methods to documentation

### LINKED ISSUE #209: Insert link to example sheet

flaviorangel:
I included the from_table methods to bearing and shaft class tutorial. I was not able to include an example code, as the Jupyter was not being able to locate the files, so I left without code examples. We may include them in the future.
Close #209


GabrielBachmann:
I've changed the plots from matplotlib to bokeh, do you think this is good to merge?

flaviorangel:
> I've changed the plots from matplotlib to bokeh, do you think this is good to merge?

It is to me, but it needs at least one review.

PR #229, opened 2019-07-30
Add test to rotor_example eigenvalues sort

### LINKED ISSUE #145: Eigenvalues sorting returning different results on appveyor
The following example was returning a different value on appveyor.
```python
        Examples
        --------
        >>> rotor = rotor_example()
        >>> evalues, evectors = rotor._eigen(0, sorted_=False)
        >>> idx = rotor._index(evalues)
        >>> idx[:6] # doctest: +ELLIPSIS
        array([ 1,  3,  5,  7,  9, 11]...
```
Results on appveyor:
```python
Expected:
    array([ 1,  3,  5,  7,  9, 11]...
Got:
    array([ 1,  3,  5,  7, 11,  9], dtype=int32)
```
We should implement this test on test_rotor_assembly.py as well.

raphaeltimbo:
The purpose of this test is to check if the index of the sorted
eigenvalues are the same on windows and linux.
Values were different when trying doctest.

Closes #145

### DIFF
--- a/ross/tests/test_rotor_assembly.py
+++ b/ross/tests/test_rotor_assembly.py
@@ -50,6 +50,14 @@ def test_index_eigenvalues_rotor1(rotor1):
     assert_almost_equal([4, 2, 0, 1, 3, 5], rotor1._index(evalues2))
 
 
+def test_evals_sorted_rotor_example():
+    rotor = rotor_example()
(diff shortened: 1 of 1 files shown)


raphaeltimbo:
Ok. For some reason, this was giving different results on doctests.
I believe that this was occurring only on python 32 and specifically 
on the doctests.
I will merge this test and close the issue for now.

PR #231, opened 2019-07-30
Add from_table class method to disk element and general from_table method

### LINKED ISSUE #226: Add .from_table method to disks
We need to implement the .from_table() method for the disks.
For now, I have submitted a PR #225 to add a small test and to modify the shaft spreadsheet so that they include a 'More' sheet with the disks information.

### LINKED ISSUE #228: Unify from_table methods
We created different from_table methods for bearings, shafts, and now disks. Except for instantiating each class, the routines are very similar, in some parts they are exact the same. It is a bad programming habit to keep different parts of the code executing the same steps, so we should unify these methods.

flaviorangel:
Close #226 and close #228


ross/utils.py
```diff
@@ -0,0 +1,237 @@
...
+                    if row[i].lower() == material_header_key_word:
+                        material_header_index = index
+                        material_header_found = True
+                        break
+            if material_header_found:
+                break
+        if not material_header_found:
+            sys.exit("Could not find the header for the materials. Make sure the table has a header "
+                     "with the parameters for the materials that will be used. There should be a column "
+                     "named " + material_header_key_word + ".")
```
raphaeltimbo: Will should we use `sys.exit` here and not `raise`?

ross/utils.py
```diff
@@ -0,0 +1,237 @@
...
+                name="shaft_mat_" + str(material_name[i]),
+                rho=material_rho[i],
+                E=material_e[i],
+                G_s=material_g_s[i],
+            )
+            new_materials["shaft_mat_" + str(material_name[i])] = new_material
+
+    if not is_csv:
+        df = pd.read_excel(file, header=header_index, sheet_name=sheet_name)
+    else:
```
raphaeltimbo: I think that with this method we are not going to need to import from .csv files, so we could remove that part of the code.

ross/utils.py
```diff
@@ -0,0 +1,237 @@
+import pandas as pd
```
raphaeltimbo: We could evaluate if we really need pandas or if we could do this using only xlrd, since pandas is a big dependency and it also adds some import time.
This is just a general comment for our evaluation in the future since we would need to remove the use of pandas also from the rotor_assembly module.

flaviorangel: We should open an issue for this one.

PR #232, opened 2019-07-30
Add Check slenderness ratio method

rodrigomoliveira1:
Adds a method that will return the colored rotor plot if the condition to the mininum slenderness ratio is not met.

Now using `.plot.rotor()` won't display the colored rotor anymore. Instead, the user will call `.check_slenderness_ratio()` to check this attribute.


ross/shaft_element.py
```diff
@@ -616,24 +616,23 @@ def patch(self, position, SR, ax):
             )
         )
 
-    def bokeh_patch(self, position, SR, check_sld, bk_ax):
+    def bokeh_patch(self, position, check_sld, bk_ax):
```
raphaeltimbo: We still have some coupling on the code here (`check_sld`), but I will merge this now and later we can test a way of decoupling this.

PR #272, opened 2019-09-04
Remove unnecessary code from freq_response()

rodrigomoliveira1:
In plot_magnitude_bokeh() there were a piece of code that only belongs to api_report.py and should not be in results.py.
I'm removing it.


raphaeltimbo:
@rodrigomoliveira1 , it looks like there are some conflicts on this PR that must be resolved before merging.

rodrigomoliveira1:
I'm going to remake this PR later.

PR #277, opened 2019-09-05
Add rotor link

### LINKED ISSUE #245: Include bearing housing dynamics
It is important to be able to include in the modelling the stiffness and inertia of the bearing housing, external supports, squeeze filme dampers or any other kind of support dynamics other than the actual oil film.

We usually have mass and stiffness parameters of these supports. So a easy way to include this in the model is create a point mass element and connect it to the rotor via the oil film sitffness and to the ground via the support stiffness.

It would be best if the software allowed the user to create multiple point mass elements and connect them to any other two nodes, that might be a rotor node, another point mass node or the ground.
This way we can model the full system including, for example, the squeeze filme damper, the bearing housing and baseplate, when needed.

raphaeltimbo:
The purpose of this PR is to add the possibility of simulating bearing housing and supports, as requested on #245.

It is still a working in progress.

- [x] Create class `PointMass`;
- [x] Add an additional argument `n_link` to bearing;
- [x] Change rotor assembly to check for links when assembling global matrices;
- [x] Add patches for drawing.

Closes #245.


ross/point_mass.py
```diff
@@ -0,0 +1,48 @@
...
+
+    Parameters
+    ----------
+    n: int
+        Node which the bearing will be located in.
+    m: float
+        Mass for the element.
+    """
+
+    def __init__(self, n=None, m=None):
```
atbrandao: I would include the option of informing two different masses, like Mx and My. In some cases the support inertia can be different in X and Y directions.
This would be quite simple to implement. Just include optional arguments `Mx` and `My` that will be used to built the matrix M. If only the argument `m` is provided, then we assign `Mx = My = m`.

raphaeltimbo: Thanks for the suggestion!
This was included on commit 6392579

raphaeltimbo:
Consireding the following code:
```python
import ross as rs

steel = rs.materials.steel

i_d = 0
o_d = 0.05
n = 6
L = [0.25 for _ in range(n)]

shaft_elem = [
    rs.ShaftElement(
        l, i_d, o_d, steel, shear_effects=True, rotary_inertia=True, gyroscopic=True
    )
    for l in L
]

disk0 = rs.DiskElement.from_geometry(
    n=2, material=steel, width=0.07, i_d=0.05, o_d=0.28
)
disk1 = rs.DiskElement.from_geometry(
    n=4, material=steel, width=0.07, i_d=0.05, o_d=0.28
)

stfx = 1e6
stfy = 0.8e6
bearing0 = rs.BearingElement(0, n_link=7, kxx=stfx, kyy=stfy, cxx=0)
support = rs.BearingElement(7, kxx=stfx, kyy=stfy, cxx=0, tag="Support")
bearing1 = rs.BearingElement(6, kxx=stfx, kyy=stfy, cxx=0)

point_mass = rs.PointMass(7, m=1.)

rotor = rs.Rotor(shaft_elem, [disk0, disk1], [bearing0, support, bearing1], [point_mass])
```
The current state of the plot is:

image

We still need to add the point mass plot. I also think we should have more space between the 
support and bearing spring/damper. But this we can treat in a different PR.

PR #290, opened 2019-09-17
WIP - Add Stability level 1 analysis

rodrigomoliveira1:
This analysis consider a anticipated cross coupling QA based on conditions at the normal operating point and the cross-coupling required to produce a zero log decrement, Q0.

### DIFF
--- a/ross/api_report
+++ b/ross/api_report
@@ -493,7 +493,18 @@ class Report:
 
         return node_min, node_max
 
-    def stability_level_1(self):
+    def stability_level_1(
+            self,
+            machine_type,
+            D,
+            H,
+            HP,
+            N,
+            RHOd,
+            RHOs,
+            steps=21,
+            unit="m",
+    ):
         """Stability analysis level 1.
 
         This analysis consider a anticipated cross coupling QA based on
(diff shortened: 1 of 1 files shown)


ross/api_report
```diff
@@ -505,10 +516,161 @@ class Report:
...
+            Bt = 1.5
+            Dt, Ht = D, H
+            Qa = 0.
+            for i in range(len(self.rotor.disk_elements)):
+                qi = (HP[i] * Bt * C) / (Dt[i] * Ht[i] * N)
+                Qa = Qa + qi
+
+        bearings = [copy(b) for b in self.rotor.bearing_seal_elements]
+        for brg in self.rotor.bearing_seal_elements:
+            cross_coupling = BearingElement(n=brg.n, kxx=0, cxx=0, kxy=Qa, kyx=-Qa)
```
raphaeltimbo: I don't think the anticipated cross-coupling should be applied to the bearing position.
Can you verify that?

ross/api_report
```diff
@@ -505,10 +516,161 @@ class Report:
...
+        rotor0 = Rotor(
+            self.rotor.shaft_elements,
+            self.rotor.disk_elements,
+            bearings,
+            self.rotor.rated_w
+        )
+        non_backward = rotor0.whirl_direction() != "Backward"
+        log_dec_a = rotor0.log_dec[non_backward][0]
+
+        stiffness = np.linspace(0, 10 * Qa, steps)
```
raphaeltimbo: `stiffness` is not a good name in this case. Maybe `cross_coupled_stiffness`...

ross/api_report
```diff
@@ -505,10 +516,161 @@ class Report:
...
+                bearings,
+                self.rotor.rated_w
+            )
+            non_backward = rotor1.whirl_direction() != "Backward"
+            log_dec[i] = rotor1.log_dec[non_backward][0]
+
+        g = interp1d(stiffness, log_dec, fill_value='extrapolate', kind="linear")
+
+        stiff = stiffness[-1] * (1 + 1 / (len(stiffness)))
+        while g(stiff) > 0:
```
raphaeltimbo: I could not understand what this loop is doing. Could you please explain?

ross/api_report
```diff
@@ -505,10 +516,161 @@ class Report:
...
+            log_dec = np.append(log_dec, g(stiff))
+            stiffness = np.append(stiffness, stiff)
+            stiff += stiffness[-1] / (len(stiffness))
+
+        # Cross-coupled required to prodeuce zero log dec - API 684 - SP6.8.5.8
```
raphaeltimbo: produce

ross/api_report
```diff
@@ -505,10 +516,161 @@ class Report:
...
+            C = 9.55
+        elif unit == "in":
+            C = 63.0
+        else:
+            raise TypeError("choose between meters (m) or inches (in)")
+
+        if len(D) != len(H):
+            raise Exception("length of D must be the same of H")
+
+        # Qa - Antecipated cross-coupling - API 684 - SP6.8.5.6
```
raphaeltimbo: Anticipated

ross/api_report
rodrigomoliveira1: You are right. I'm fixing it right now and making some other improvements. It's based on if the rotor is between bearings or overhung, and the cross-coupling is introduced at rotor's mid-span or at the center of gravity of the stage or impeller for single overhung rotors.

ross/api_report
rodrigomoliveira1: Noted!

ross/api_report
rodrigomoliveira1: SP6.8.5.8. b. - The amount of the applied cross coupling required to produce a zero log decrement, Q0. This value can be reached by extrapolation or linear interpolation between two adjacent points on the curve.
So, basically I'm doing a loop to extrapolate the cross-coupled until the log dec reaches zero.

ross/api_report
rodrigomoliveira1: Noted!

ross/api_report
rodrigomoliveira1: Noted!

ross/api_report
raphaeltimbo: For that I think it would be better to use [newton method](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.newton.html).

PR #319, opened 2019-10-01
Add .save() and .load() to point mass element

rodrigomoliveira1:
In PR #318, I should have added these methods too.
So, I'm adding them now.
Now, the code will save and load the point mass elements.

### DIFF
--- a/ross/point_mass.py
+++ b/ross/point_mass.py
@@ -101,6 +101,61 @@ class PointMass(Element):
             f" my={self.my:{0}.{5}}, tag={self.tag!r})"
         )
 
+    def save(self, file_name):
+        """Saves a point mass element in a toml format. It works as an
+        auxiliary function of the save function in the Rotor class.
+        Parameters
+        ----------
+        file_name: string
+            The name of the file the point mass element will be saved in.
+
+        Returns
+        -------
+        None
+
+        Examples
+        --------
+        >>> pointmass = point_mass_example()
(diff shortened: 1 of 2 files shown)


rodrigomoliveira1:
While updating the save and load methods, I have found out that, when loading a rotor, the rotor file must have a toml for every type of element, otherwise it will raise an error because it can't find a "PointMass.toml" or "DiskElement.toml", etc...

When saving a rotor, the code will only build toml files for the element types presented in the model. That is, if a rotor is built with shaft and bearing elements, the rotor file gets only "ShaftElement.toml" and "BearingElement.toml". If we try to load it later, we get an "FileNotFound" error.

raphaeltimbo:
Instead of `os.chdir` we should use `pathlib.Path()`.
It is not a good idea to change the running directory since the user is not expecting that and we always have to worry that we are going back to the directory we were before. This is error prone.
Using `Path()` give us the ability to use relative paths and it is much easier to use.
Regarding the `try` and `except` statements, if we use `Path()` we can use the `.glob()` method to see the files that we have inside the folder and then we can iterate on them creating the desired objects.

raphaeltimbo:
@rodrigomoliveira1, I believe this is being implemented in a different PR, right?
Can we close this one?

rodrigomoliveira1:
> @rodrigomoliveira1, I believe this is being implemented in a different PR, right?
> Can we close this one?
Yes, PR #360 replaces this PR. We can close this one.

PR #337, opened 2019-10-07
Add ModalResults

raphaeltimbo:
All the .run methods that we have return a Results object that holds
some attributes such as arrays and has some plot methods. The only
exception to that is our run_modal method, which returns None and
change some rotor attributes in place (such as evalues, evectors,
wn etc.). This different behavior is causing confusion in some new
users when they are going through the tutorial. To solve this,
I am creating a ModalResults class that will have the speed as the main
argument, and the returned object will hold the attributes that were
stored on the rotor before.


raphaeltimbo:
I am investigating why the following doctest is failing:
```python
    >>> from ross.fluid_flow.fluid_flow import pressure_matrix_example
    >>> my_fluid_flow = pressure_matrix_example()
    >>> ax = matplot_eccentricity(my_fluid_flow, z=int(my_fluid_flow.nz/2))
    >>> # to show the plots you can use:
    >>> # plt.show()
```
It is giving the following error:
```python
UNEXPECTED EXCEPTION: TypeError("object of type 'numpy.float64' has no len()")
Traceback (most recent call last):
  File "/opt/python/3.7.1/lib/python3.7/doctest.py", line 1329, in __run
    compileflags, 1), test.globs)
  File "<doctest ross.fluid_flow.fluid_flow_graphics.matplot_eccentricity[2]>", line 1, in <module>
  File "/home/travis/build/ross-rotordynamics/ross/ross/fluid_flow/fluid_flow_graphics.py", line 202, in matplot_eccentricity
    ax.plot(pressure_matrix_object.xre[z][j], pressure_matrix_object.yre[z][j], "r.")
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/mpl_toolkits/mplot3d/axes3d.py", line 1538, in plot
    zs = np.broadcast_to(zs, len(xs))
TypeError: object of type 'numpy.float64' has no len()
```

There are two things that I still cannot understand:
1. The failure only occurs when we run the doctests in all files, if we run `pytest fluid_flow/fluid_flow_graphics.py` the test will pass;
2. The exception is raised on the `axes3d.py` modules, and we are not using a 3d plot here.

I believe that the changes on results.py may have some implication here, since this is the only module were wi import Axes3d.

raphaeltimbo:
The problem was the following:
The last plot example before fluid_flow_graphics was this:
```python
        Example
        -------
        >>> rotor = rotor_example()
        >>> modal = rotor.run_modal(speed=0)
        >>> modal.wn[:2]
        array([91.79655318, 96.28899977])
        >>> modal.wd[:2]
        array([91.79655318, 96.28899977])
        >>> modal.plot_mode(0) # doctest: +ELLIPSIS
        (<Figure ...
```
This one uses the `ax.plot(projection='3d')`. When we started the matplot_eccentricity test we go through these lines:
```python
if ax is None:
    ax = plt.gca() # (get the current axes)
```
Since we were not providing an axes in the example, the code would try to use the current axes which had a 3d projection and this would cause the error reported above.

We can eliminate this problem by changing the example to:
```python
    >>> import matplotlib.pyplot as plt
    >>> from ross.fluid_flow.fluid_flow import pressure_matrix_example
    >>> my_fluid_flow = pressure_matrix_example()
    >>> fig, ax = plt.subplots()
    >>> ax = matplot_eccentricity(my_fluid_flow, z=int(my_fluid_flow.nz/2), ax=ax)
    >>> # to show the plots you can use:
    >>> # plt.show()
```

PR #344, opened 2019-10-14
Fix forced plots

raphaeltimbo:
This fixes some issues on the force plots such as figure size and
labels.

### DIFF
--- a/ross/results.py
+++ b/ross/results.py
@@ -984,7 +984,7 @@ class FrequencyResponseResults:
         # bokeh plot - create a new plot
         mag_plot = figure(
             tools="pan, box_zoom, wheel_zoom, reset, save",
-            width=600,
+            width=640,
             height=240,
             title="Frequency Response - Magnitude",
             x_axis_label="Frequency",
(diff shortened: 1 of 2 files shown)


ross/results.py
```diff
@@ -1360,15 +1360,15 @@ def plot_magnitude_bokeh(self, dof, units="m", **kwargs):
...
             mag = 2 * mag * 1e6
-            y_axis_label = "Amplitude $(\mu pk-pk)$"
+            y_axis_label = "Amplitude (μ pk-pk)"
 
         # bokeh plot - create a new plot
         mag_plot = figure(
             tools="pan, box_zoom, wheel_zoom, reset, save",
-            width=900,
-            height=400,
+            width=600,
```
rodrigomoliveira1: The standard width is 640px on other plots. Did you need to reduce it in this plot?

ross/results.py
```diff
@@ -1449,9 +1449,9 @@ def plot_phase_bokeh(self, dof, **kwargs):
 
         phase_plot = figure(
             tools="pan, box_zoom, wheel_zoom, reset, save",
-            width=900,
-            height=400,
-            title="Forced Response - Magnitude",
+            width=600,
```
rodrigomoliveira1: same

ross/results.py
raphaeltimbo: I will correct this.

PR #346, opened 2019-10-15
Add __init__.py file

rodrigomoliveira1:
Add `__init__.py` file to include rotors folder when installing ross


rodrigomoliveira1:
I'm canceling this PR because I need to find better idea for how to fix the installing rotors' folder problem.

PR #360, opened 2019-10-28
Rotor save and load new format

GabrielBachmann:
This PR changes the way we save and load rotors.
The new format does not use [os.chdir](https://docs.python.org/2/library/os.html#os.chdir) in Rotor class.


ross-bott:
Hi there!
less than 45 days;

GabrielBachmann:
#432 will substitute this PR

PR #366, opened 2019-11-01
Improvements to api_report.py - Add test_api_report.py

rodrigomoliveira1:
- Add Stability level 1 screening criteria;
- Modifies the code to storage the cross-coupling range for each rotor stage (or impeller). This is useful to distinguish the cross-coupling evaluation for different `rotor_types` (between bearings, overhung...);
- Add `.summary()` method;
- Add summary for stability level 1


rodrigomoliveira1:
I have facing a strange behavior from the CI's in this PR. 
First of all, Travis was returning an error testing on Python 3.7 when running a `numpy.linspace()` command. Well, I just changed what I was doing and fixed the problem. However, now. the Appveyor is point an error while running test_fluid_flow.py, which does not make any sense because it has not been changed at all.

PR #376, opened 2019-11-12
Visual improvements to graphs

rodrigomoliveira1:
The labels, ticks and titles font size from bokeh figure were small when putting it presentations. So I've increased the font size of all graphs.

- Add new tab to the `.summary()` with stability level 2 data;
- Change stability level 2 labels for better user understanding when using `.summary()`. The labels include explicitly all the components considered in a certain analysis (Shaft + Bearing + Disk + Seal...);
- Increases labels, titiles and tick font sizes (results.py and api_report.py);
- Add mcs speed to evaluate mode shapes (api_report.py);
- Fix `.unbalance_response()` plot size to match results.py file;
- In `.stability_level_1()` remove condition from returning and add it as attribute;
- `.run_modal().plot_mode()` add legend informing the whirl direction.

### DIFF
--- a/ross/api_report.py
+++ b/ross/api_report.py
@@ -1069,9 +1069,9 @@ class Report:
             log_dec_disk.append(modal.log_dec[non_backward][0])
 
         # Evaluate log dec for group bearings + disks
-        disk_tags = [disk.tag for disk in self.rotor.disk_elements]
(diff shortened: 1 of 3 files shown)


rodrigomoliveira1:
Once again we're experiencing an appveyor failure only.
I'm removing the test which is returning an error.

image

Issue #382, opened 2019-11-18
Links are not working for more than one support


raphaeltimbo:
The current test where we have links has only one support.
We should add another test with supports on both bearings to evaluate if the implementation is correct.

rodrigomoliveira1:
I have tried this simple example below and it doesn't work:

```
def rotor_example():
    i_d = 0
    o_d = 0.05
    n = 6
    L = [0.25 for _ in range(n)]

    shaft_elem = [
        ShaftElement(
            l, i_d, o_d, steel, shear_effects=True, rotary_inertia=True, gyroscopic=True
        )
        for l in L
    ]

    disk0 = DiskElement.from_geometry(
        n=2, material=steel, width=0.07, i_d=0.05, o_d=0.28
    )
    disk1 = DiskElement.from_geometry(
        n=4, material=steel, width=0.07, i_d=0.05, o_d=0.28
    )

    stfx = 1e6
    stfy = 0.8e6
    bearing0 = BearingElement(0, n_link=7, kxx=stfx, kyy=stfy, cxx=0, scale_factor=1)
    support0 = BearingElement(7, kxx=stfx, kyy=stfy, cxx=0, tag="Support", scale_factor=1)
    bearing1 = BearingElement(6, n_link=8, kxx=stfx, kyy=stfy, cxx=0, scale_factor=1)
    support1 = BearingElement(8, kxx=stfx, kyy=stfy, cxx=0, tag="Support", scale_factor=1)

    point_mass0 = PointMass(7, m=1.0)
    point_mass1 = PointMass(8, m=1.0)

    return Rotor(shaft_elem, [disk0, disk1], [bearing0, support0, support1, bearing1],
                 [point_mass0, point_mass1])

rotor = rotor_example()
rotor.M().shape

Traceback (most recent call last):

  File "<ipython-input-41-d8ae27a43e7d>", line 1, in <module>
    rotor.M().shape

  File "C:/Users/USER/Google Drive/Backup notebook/Lavi/ross-master/ross/rotor_assembly.py", line 678, in M
    M0[np.ix_(dofs, dofs)] += elm.M()

IndexError: index 32 is out of bounds for axis 0 with size 32
```
The size of the mass matrix is correct (32 x 32), however when we come to observe the elements' indexes we can notice a bug on the last line (shown in the print below). The Global Index should be `GlobalIndex(x_8=30, y_8=31)`

```
[...]

PR #407, opened 2019-12-18
Fix links not working for more than one support

rodrigomoliveira1:
Related to Issue #382.

We were having problems to define de global indexes for the elements using this method intenally to a element class.
It makes more sense to use .dof_global_index() when build the rotor model.
The routine to calculate the elements' global indexes have been moved to rotor_assembly (`.__init__()` method).
Now, each element has an attribute that keeps those index, instead of calling a function to this job.

### DIFF
--- a/ross/rotor_assembly.py
+++ b/ross/rotor_assembly.py
@@ -1052,7 +1052,8 @@ class Rotor(object):
         array([[[1.00000000e-06, 1.00261725e-06, 1.01076952e-06, ...
         """
         if speed_range is None:
-            speed_range = np.linspace(0, max(self.evalues.imag) * 1.5, 1000)
+            modal = self.run_modal(0)
+            speed_range = np.linspace(0, max(modal.evalues.imag) * 1.5, 1000)
 
         freq_resp = np.empty((self.ndof, self.ndof, len(speed_range)), dtype=np.complex)
 
--- a/ross/rotor_assembly.py
+++ b/ross/rotor_assembly.py
@@ -368,8 +368,8 @@ class Rotor(object):
         )
 
         # global indexes for dofs
+        n_last = self.shaft_elements[-1].n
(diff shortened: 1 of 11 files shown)


ross/rotor_assembly.py
```diff
@@ -365,6 +367,44 @@ def flatten(l):
             + 2 * len([el for el in point_mass_elements])
         )
 
+        # global indexes for dofs
+        for elm in self.elements:
```
raphaeltimbo: I think we can improve this part of the code doing something like:
```python
n = 0
for elm in self.elements:
    # get the local_index tuple
    local_index = elm.dof_local_index()
    # get the global index by iterating on the local_index and adding n
    global_index = dof_tuple({k: v + n for k, v in local_index._asdict().items()})
    # set n as the last value from the global index
    n = global_index[-1]
```
Also, we should avoid checking types with `if elm.__class__.__name__ == '<type>'`. 
It is better to use `isinstance()` if necessary. For example, for the bearing case you would have `isinstance(elm, BearingElement)` and this would work for all types described in `bearing_class_list`, since they all inherit from `BearingElement`. This way, you would not need to update this list everytime a new bearing type is added.

But I think the type checking can all be avoided in this case, since we only need `n` from the last global index and the local index for the current element.

rodrigomoliveira1: I understood the isistance() part and I've modified it. But I didn't get the code suggestion right.
Should I replace all the code lines inside the `for` loop (line 370) by the code you wrote or just add these lines? I'm asking because `dof_tuple` would be referenced before assignment this way.
```
n = 0
for elm in self.elements:
    local_index = elm.dof_local_index()

    # in the line below dof_tuple is not defined
    global_index = dof_tuple({k: v + n for k, v in local_index._asdict().items()})

    n = global_index[-1]
```

PR #427, opened 2020-01-21
PDF Report preview

GabrielBachmann:
ROSS PDF ROTORDYNAMICS REPORT
===
This PR includes a preview of how the PDF generating code will function.
The folder API_Report contains a .py archive named Report.py. 
When executed, this archive :
* instantiates an example rotor;
* saves an svg file containing the plot_rotor() output;
* generates a PDF with all the data gathered.

Observations
==

I struggled trying to use Bootstrap. Therefore I've decided not to use it.
Nevertheless, after some further investigation, I've discovered the font and color pallet used on the Lumen theme.

What's next?
==
* This API_Report folder will not exist initially. Being generated inside a rotor saving folder for every rotor.
* In this PDF I will implement the tests assembled by @rodrigomoliveira1 .


ross-bott:
Hi there!
less than 45 days;

Issue #362, opened 2019-10-29
ShaftElement - two methods to load data


raphaeltimbo:
In the `ShaftElement` class, we have two methods to load data:
- `.load()`
- `.load_data()` (inherited from `Element`)

We need to remove one of these methods.

ross-bott:
Hi there!
I have marked this issue as stale because it has not had activity for 45 days.
Consider the following options:
- If the issue refers to a large task, break it in smaller issues that can be solved in
less than 45 days;
- Label the issue as `wontfix` or `wontfix for now` and close it.

rodrigomoliveira1:
PR #360 fixed this issue.

Issue #341, opened 2019-10-11
Folder "rotors" does not appear when installing ROSS


rodrigomoliveira1:
Yesterday I was testing the code during a class and when people installed ROSS, the folder "rotors" (which keeps the saved rotor models) was not being installed. Everything was running fine, but we were unable to save the files, only with we had created the folder manually.

rodrigomoliveira1:
I think we can solve this problem adding /rotors diretory to Manifest.in, as it's explained in this [PyPI tutorial](https://packaging.python.org/guides/using-manifest-in/)

ross-bott:
Hi there!
I have marked this issue as stale because it has not had activity for 45 days.
Consider the following options:
- If the issue refers to a large task, break it in smaller issues that can be solved in
less than 45 days;
- Label the issue as `wontfix` or `wontfix for now` and close it.

raphaeltimbo:
@rodrigomoliveira1 will try to solve this using the MANIFEST.in file.

GabrielBachmann:
What about giving the option for the user to give a path setting the default value to the rotors folder, then checking if the folder exists(creating a folder if not), as the code bellow:

```python
def save(self, file_name, path_rotors=Path(os.path.dirname(ross.__file__))/'rotors'):
        if not path_rotors.is_dir() and path_rotors.suffix == '':
            path_rotors.mkdir()
```

raphaeltimbo:
@GabrielBachmann , we wanted to have the same behavior that we have in other Python libraries when we try to save a file.
For example, in pandas:
```python
>>> import pandas as pd
>>> df = pd.DataFrame({’name’: [‘Raphael’, ‘Donatello’],
... ‘mask’: [‘red’, ‘purple’],
... ‘weapon’: [‘sai’, ‘bo staff’]})
```
The following will save a `a.csv` file to the current working directory.
```python
>>> df.to_csv(path_or_buf=’a.csv’)
```
The following will raise an exception, since the `test/` directory does not exist:
```python
>>> df.to_csv(path_or_buf=’test/a.csv’)
FileNotFoundError: [Errno 2] No such file or directory: ‘test/a.csv’
```

So, if the user wants to save to a folder he needs to declare that in the ‘path’ argument. Having a default folder different from the current working directory is not the expected behavior for a Python library.

rodrigomoliveira1:
We changed how we save rotors model. The user now inputs where it should be saved.
The folder "rotors" is not necessary anymore.

Issue #293, opened 2019-09-18
Material code is unreachable


raphaeltimbo:
In some `Material` methods we have the following code:
```python
def use_material(name):
    run_path = os.getcwd()
    ross_path = os.path.dirname(rs.__file__)
    os.chdir(ross_path)

    data = Material.load_data()
    try:
        material = data["Materials"][name]
        return Material(**material)
    except KeyError:
        raise KeyError("There isn't a instanced material with this name.")
    os.chdir(run_path)
```
The problem here is that the `os.chdir(run_path)` command will never run. In the try except statement the code will either suceed and return or it will raise a keyerror, never reaching the `os.chdir` command.

raphaeltimbo:
One option would be to change the way we call the material methods.
Instead of changing the directory with `os.chdir`, we could give the complete path for the file:
```python
def available_materials():
    ross_path = Path(rs.__file__).parent / "available_materials.toml"

    try:
        data = Material.load_data(ross_path)
        return list(data["Materials"].keys())
    except FileNotFoundError:
        return "There are no saved materials."
```
The `load()` method would be something like:

```python
    def load_data(file_path):
        try:
            with open(file_path, "r") as f:
                data = toml.load(f)
        except FileNotFoundError:
            data = {"Materials": {}}
            Material.dump_data(data)
        return data
```
This way we would avoid changing directories back and forth.

ross-bott:
Hi there!
I have marked this issue as stale because it has not had activity for 45 days.
Consider the following options:
- If the issue refers to a large task, break it in smaller issues that can be solved in
less than 45 days;
- Label the issue as `wontfix` or `wontfix for now` and close it.

raphaeltimbo:
Solved by #432

PR #432, opened 2020-02-04
Updating PR 360 and other small changes

GabrielBachmann:
This PR's purpose is to update PR #360 and to solve the following issues:
*  Folder "rotors" does not appear when installing ROSS #341  
* ShaftElement - two methods to load data #362 
*  Material code is unreachable #293.  

Removing completely the rotors folder from source.
Most of the changes were just code refactoring(changing all the load_data() methods to get_data() for a better understanding for the user and changing the variable file_name to file_path) and removing the usage of os.chdir() from the code.
Besides that, I've noticed a mistake in the __eq__ method of material class, where the output of that function was being always True.


ross/rotor_assembly.py
```diff
@@ -74,7 +75,7 @@ class Rotor(object):
         List with the shaft elements
     disk_elements : list
         List with the disk elements
-    bearing_seal_elements : list
+    bearing_elements : list
```
rodrigomoliveira1: Same question made in PR #433 about this changing.

ross/bearing_seal_element.py
```diff
@@ -307,7 +310,7 @@ def __eq__(self, other):
     def __hash__(self):
         return hash(self.tag)
 
-    def save(self, file_name):
+    def save(self, file_name=Path(os.getcwd())):
```
raphaeltimbo: If we have `file_name` with a default value, the user can create two different bearings and when saving we would overwrite the first bearing. Maybe we `file_name` should be a required positional argument?

ross/api_report.py
```diff
@@ -805,7 +806,7 @@ def stability_level_1(self, D, H, HP, oper_speed, RHO_ratio, RHOs, RHOd, unit="m
         # remove disks and seals from the rotor model
         bearing_list = [
```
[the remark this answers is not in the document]
GabrielBachmann: I think that's a good approach too, thanks!

GabrielBachmann:
I don't know why, but I'm getting this error log: 
```Python
=================================== FAILURES ===================================
____________________________ test_stability_level2 _____________________________
report0 = <ross.api_report.Report object at 0x7f19513ab470>
report1 = <ross.api_report.Report object at 0x7f195101f128>
report2 = <ross.api_report.Report object at 0x7f194eebd518>
    def test_stability_level2(report0, report1, report2):
        df0 = report0.stability_level_2()
        df1 = report1.stability_level_2()
    
        assert_allclose(
            df0["log_dec"].tolist(),
            [
                0.17669652215696618,
                0.15801304519741688,
                0.14761936907792816,
                0.3153250139555406,
                0.1476193691000094,
            ],
>           atol=1e-6,
        )
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=1e-06
E       
E       (shapes (7,), (5,) mismatch)
E        x: array([0.176697, 0.158013, 0.147619, 0.285484, 0.285484, 0.315325,
E              0.147619])
E        y: array([0.176697, 0.158013, 0.147619, 0.315325, 0.147619])
tests/test_api_report.py:255: AssertionError
```

rodrigomoliveira1:
Do you get this error running it localy?

GabrielBachmann:
Yes, I get the same error message !

ross/bearing_seal_element.py
raphaeltimbo: Sorry, I think I was not clear in my comment.
The previous code had `file_name` as a required positional argument, so if the user tried to call `.save()` without providing the `file_name` this would raise an exception, which is the expected behavior.
With the current code (`def save(self, file_name=Path(os.getcwd())):`, having `file_name` as a keyword argument with a default value, the user can call `.save()` without the `file_name` argument and it will save the file, overwriting anything that is already there, which is a behavior that the user does not expect (in other libraries `.save` methods you have to provide a path).
So, I was actually asking that `.save` methods had the signature `def save(file_name)`, instead of `def save(file_name=Path(os.getcwd()))`.

GabrielBachmann: Oh my mistake, I understand this now!

GabrielBachmann: Currently the code is looking like this:
```Python
 def save(self, file_name):
        """Saves a bearing element in a toml format. It works as an auxiliary
        function of the save function in the Rotor class.
        Parameters
        ----------
        file_name: string
            The name of the file the bearing element will be saved in.
        Returns
        -------
        None
        Examples
        --------
        >>> bearing = bearing_example()
        >>> bearing.save(Path(os.getcwd()))
        """
        data = self.get_data(Path(file_name)/'BearingElement.toml')
``` 
The code already complies with the specifications!

PR #436, opened 2020-02-07
Add tapared tutorial

### LINKED ISSUE #416: Improving the tutorial - ShaftElement
>How to create a shaft element with different outer diameter and length should be shown in this tutorial. We should keep in mind that the users are not expert in Python.
L = 0.25
i_d = 0
o_d = 0.05
N = 6       # Number of elements
l_list = [L for _ in range(N)]
shaft_elements = [
    rs.ShaftElement(
        L=l,
        i_d=i_d,
        o_d=o_d,
        material=Steel,
        shear_effects=True,
        rotary_inertia=True,
        gyroscopic=True,
    )
    for l in l_list
]

raphaeltimbo:
This adds the ShaftTaperedElement to the tutorial so that the docs will
be synced with the released 0.2.0 version.

Closes #416.


rodrigomoliveira1:
There's some kind of conflict here, I think you used the same branch than before to open this new PR, and now it's bringing along the same commits done on your last PR about the Tutorial.

raphaeltimbo:
> There's some kind of conflict here, I think you used the same branch than before to open this new PR, and now it's bringing along the same commits done on your last PR about the Tutorial.

I will open a new PR to fix this.

Issue #437, opened 2020-02-10
Decrease the number of functions in fluid_flow_coefficients.py.


JuliaMota:
In fluid_flow, the same function can return the numerical solution or some approximation of the literature according to the user's choice. But for the calculation of the stiffness and damping coefficients there are different functions for each case. Correcting this, it is possible to decrease the number of functions in fluid_flow_coefficients.py.

PR #438, opened 2020-02-10
Organizes coefficients functions

JuliaMota:
This pull request solves the issue #437


ross/fluid_flow/fluid_flow_coefficients.py
```diff
@@ -182,14 +93,17 @@ def calculate_oil_film_force(fluid_flow_object, force_type=None):
...
     Parameters
     ----------
     fluid_flow_object: A FluidFlow object.
     oil_film_force: str
-        If set, calculates the oil film force analytically considering the chosen type: 'short' or 'long'.
+        If set, calculates the oil film force analytically analytically considering the chosen type: 'short'.
         If set to 'numerical', calculates the oil film force numerically.
+    force_type: str
+        If set, calculates the stiffness matrix analytically considering the chosen type: 'short'.
+        If set to 'numerical', calculates the stiffness matrix numerically.
```
rodrigomoliveira1: I didn't understand well these docstrings.
The options to `force_type` attribute are "short" and "numerical"? The same comment goes to `oir_film_force`.
Maybe we could set `force_type` with a default argument

Also, you could write the input options, like: 
```
"""
Options are: "short" and "long".
Default is "short"
"""
``` 

At line 102, analytically is repeated.

ross/fluid_flow/fluid_flow_coefficients.py
```diff
@@ -201,30 +115,93 @@ def calculate_stiffness_matrix(fluid_flow_object, oil_film_force='numerical'):
...
 
-    return [k_xx, k_xy, k_yx, k_yy]
+def calculate_damping_matrix(fluid_flow_object, force_type=None):
+    """Returns the damping matrix calculated analytically.
+    Suitable only for short bearings.
+    Parameters
+    -------
+    force_type: str
+        If set, calculates the stiffness matrix analytically considering the chosen type: 'short'.
+        If set to 'numerical', calculates the stiffness matrix numerically.
```
rodrigomoliveira1: The same comment as before applies here.

ross/fluid_flow/fluid_flow_coefficients.py
JuliaMota: The approximations presented by Friswell for the stiffness and damping matrices are only for the case of the short bearing. So the options for force_type attribute are "short" and "numerical".

For oil_film_force we really have both cases. It was a mistake. I will correct that.

JuliaMota:
Tests are passing on my computer, but here there are problems. @rodrigomoliveira1 , can you help me with this?

rodrigomoliveira1:
> 
> 
> Tests are passing on my computer, but here there are problems. @rodrigomoliveira1 , can you help me with this?

Travis points a failure in this doctest below. I believe the signs are wrong.

image

PR #441, opened 2020-02-11
Add units

### LINKED ISSUE #440: Add option for units
We need to implement options for choosing units.
This should be available for inputting data and also for getting results (e.g. plots with speed units in RPM instead of rad/s).
An option to implement this is to use the [pint library](https://pint.readthedocs.io/en/0.10.1/).

raphaeltimbo:
If we use the check_units decorator in a function the arguments are checked, and if they are in the units
dictionary, they are converted to the 'default' unit given in the dictionary.

For example:
```python
>>> units = {
... "L": "meter",
... }

>>> @check_units
... def foo(L=None):
...     print(L)
```

If we call the function with the argument as a float:
```python
>>> foo(L=0.5)
0.5 meter
```

the L argument will be available inside the function as a pint.Quantity object
(with the magnitude and units attributes).

If we call the function with a pint.Quantity object the value is automatically converted to the default:
```python
>>> foo(L=Q_(0.5, 'inches'))
0.0127 meter
```

Closes #440

### DIFF
--- a/ross/units.py
+++ b/ross/units.py
@@ -16,12 +16,14 @@ __all__ = ["Q_", "check_units"]
 units = {
     "E": "N/m**2",
     "G_s": "N/m**2",
-    "rho": "N/m**3",
+    "rho": "kg/m**3",
     "L": "meter",
     "idl": "meter",
     "idr": "meter",
     "odl": "meter",
     "odr": "meter",
+    "speed": "radian/second",
+    "frequency": "radian/second",
 }
 
 
--- a/ross/units.py
+++ b/ross/units.py
@@ -1,0 +1,82 @@
+"""
+This module deals with units conversion in the ROSS library. 
+"""
+import inspect
+from pint import UnitRegistry
+from functools import wraps
+from pathlib import Path
+
+new_units_path = Path(__file__).parent / "new_units.txt"
+ureg = UnitRegistry()
+ureg.load_definitions(str(new_units_path))
+Q_ = ureg.Quantity
+
+__all__ = ["Q_", "check_units"]
+
+units = {
+    "E": "N/m**2",
+    "G_s": "N/m**2",
+    "rho": "N/m**3",
+    "L": "meter",
+    "idl": "meter",
+    "idr": "meter",
+    "odl": "meter",
+    "odr": "meter",
+}
+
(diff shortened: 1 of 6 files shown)


raphaeltimbo:
@rodrigomoliveira1 , I have finished this module. Could you please review it?.
For now, I have only tested on the ShaftElement, but if we merge this we can start to add units to other function calls.

ross/units.py
```diff
@@ -0,0 +1,84 @@
...
+    "G_s": "N/m**2",
+    "rho": "kg/m**3",
+    "L": "meter",
+    "idl": "meter",
+    "idr": "meter",
+    "odl": "meter",
+    "odr": "meter",
+    "speed": "radian/second",
+    "frequency": "radian/second",
+}
```
rodrigomoliveira1: Just to understand. Here we set the default units for each attribute, right? For now, it got only the ShaftElement attributes as you said.

And if the user wants to enter a unit like "feet" you just write like `L.feet` and the pint packages detects the unit "feet"?
I imagine it has a registry of several units to convert.

raphaeltimbo: With pint you can create and interact with an object called pint.Quantity (Q_) like this:
```python
# create the object
>>> x = Q_(1, 'meter')
>>> x
1 meter

# get quantity magnitude
>>> x.m # or x.magnitude
1

# get the units
>>> x.units
meter

# convert to foot
>>> x.to('foot')
3.2808398950131235 foot
```

So, if the user wants to create a ShaftElement object using foot in the L argument it would be something like this:
```python
>>> s0 = rs.ShaftElement(L=Q_(2, 'foot'), idl=...)

# The value is converted to our default units during instantiation
>>> s0.L
0.6095999999999999
```

If we give it a float the behavior is the same as before
```python
>>> s0 = rs.ShaftElement(L=2., idl=...)
>>> s0.L
2.
```

> I imagine it has a registry of several units to convert.

Yes. [Here](https://github.com/hgrecco/pint/blob/2503f00762d1bc159ab196e2de93963f257629dd/pint/default_en.txt) you can check the units available.

rodrigomoliveira1: Got it! thanks you, @raphaeltimbo!

PR #450, opened 2020-02-20
Add CoAxialRotor class

rodrigomoliveira1:
- Add CoAxialRotor class
- Add new example to run CoAxialRotor class
- Adapt `.plot_rotor()` method to plot co axial rotor system.
- Add ["shaft_number"] column to the dataframe to identify each shaft in the rotor system.
- Run Black to format code.

I still need to add some initial tests, so I'm leaving it in WIP, while you can check it and comment.

### DIFF
--- a/ross/tests/test_rotor_assembly.py
+++ b/ross/tests/test_rotor_assembly.py
@@ -851,6 +851,95 @@ def test_static_analysis_rotor5(rotor5):
 
 @pytest.fixture
 def rotor6():
+    #  Overhung rotor without damping with 10 shaft elements
+    #  2 disks and 2 bearings
+    i_d = 0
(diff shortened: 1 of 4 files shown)


ross/tests/test_rotor_assembly.py
```diff
@@ -851,91 +851,72 @@ def test_static_analysis_rotor5(rotor5):
 
 @pytest.fixture
 def rotor6():
-    #  Overhung rotor without damping with 10 shaft elements
```
raphaeltimbo: Maybe we should keep this test and add an additional test for the coaxial rotor.

rodrigomoliveira1: Done.

PR #452, opened 2020-02-20
It was added the stochastic part of the model

LuanDaSilva:
This function allows the use of vector as a parameter of function/object. It reads the vector(s) and generate a proper input to the function/object returning a list of function/objects outputs evalueted at each vector(s) value.


ross/utils.py
```diff
@@ -4,6 +4,23 @@
 from bokeh.models import LogColorMapper
 from bokeh.palettes import Viridis256
 from bokeh.plotting import figure
+import itertools
+from collections import Iterable
+
+from ross.materials import Material
```
raphaeltimbo: Is this import being used?

ross/utils.py
```diff
@@ -4,6 +4,23 @@
...
+
+def stochastic_ross(f,*args):
```
raphaeltimbo: Please add the function documentation. Also check if we could improve the name of this function to give a better idea of its purpose to the user.

rodrigomoliveira1:
I don't why I getting an error on `test_fluid_flow.py` in this PR. Once again we're experiencing strange behaviors from the CI's. I'm trying small commits to see if it vanishes eventually like the last time.

rodrigomoliveira1:
PR #478 moves the method once added in utils.py to new folder.
So, I'm closing this PR.

PR #453, opened 2020-02-21
Material units

### LINKED ISSUE #445: Material accepts inconsistent arguments
Material accepts inconsistent arguments during instantiation. See example below.

```python
>>> steel = rs.Material(name="Steel", rho=7810, E=211e9, G_s=81.2e9)
>>> steel.Poisson
0.2992610837438423
>>> steel = rs.Material(name="Steel", rho=7810, E=211e9, G_s=81.2e9, Poisson=0.2)
>>> steel.Poisson
0.2
```

raphaeltimbo:
This adds check_units to the `Material` class:

```python
import ross as rs

Q_ = rs.units.Q_
steel = rs.Material(name='steel', rho=7810, E=Q_(3e7, 'psi'), G_s=Q_(1e7, 'psi'))
```

It also closes #445


ross/materials.py
```diff
@@ -47,21 +47,28 @@ class Material:
...
+            raise ValueError("Spaces are not allowed in Material name")
+
+        given_args = []
+        for arg in ["E", "G_s", "Poisson"]:
+            if locals()[arg] is not None:
+                given_args.append(arg)
+        if len(given_args) != 2:
+            raise ValueError(
+                "At least 2 arguments from E, G_s" "and Poisson should be provided "
+            )
```
rodrigomoliveira1: This error message could lead for a wrong conclusion. When using "at least" the user could think we could input two or all the three parameters.
Maybe we could say something more restrictive like:
 ```
raise ValueError(
    "Only 2 arguments from E, G_s and Poisson should be provided."
)
```

raphaeltimbo: You are right. I will change this and update the PR.

rodrigomoliveira1:
There's still a conflict, I can't merge it.

raphaeltimbo:
> 
> 
> There's still a conflict, I can't merge it.

I will try to solve this.

PR #459, opened 2020-03-05
Change how docs are updated

raphaeltimbo:
This PR changes the process to update our documentation.

Following [semantic versioning](https://semver.org/) we will have branches that follow the current minor version (e.g. 0.3, 0.4 etc.). When we merge some modification to the docs, Travis CI will run the `deploy_docs.sh` script, which will check if the modification is being made to the branch that refers to the current released version. If that is the case, Travis will build the documentation and push the updated docs to the ross-website repo.

This avoids updating the documentation on every push to the master branch which can cause inconsistencies between the released version and the website docs.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -34,8 +34,17 @@ Documentation
 -------------
 We use `sphinx <http://www.sphinx-doc.org/en/master/>`_ to generate the project's documentation. We keep the source
 files at ~/ross/docs, and we keep the html files used to build the website in a
-`separate repository <https://github.com/ross-rotordynamics/ross-website>`_. If you want to modify the documentation,
-modify the source files and them do the following to modify the website:
+`separate repository <https://github.com/ross-rotordynamics/ross-website>`_.
+The website tracks the documentation for the released version with the following procedure:
+
+ 1. Travis runs the deploy_docs.sh file ('after_success' phase);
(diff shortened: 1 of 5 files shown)


rodrigomoliveira1:
This PR has the same problem than the last one regarding the test_fluid_flow.py file.

raphaeltimbo:
The problem is related to an update on matplotlib version 3.2.0.
You can see here in [this commit](https://github.com/matplotlib/matplotlib/commit/2cc8e14aa1918906cf45a8cdd57f452f2c05e7a4).
We will have to remove the `ax.axis("equal")`.

rodrigomoliveira1:
I see. They released a new version two days ago.

raphaeltimbo:
The problem is that matplotlib should raise this error for a polar plot, which is not the case here.
I tried to run the `test_fluid_flow.py` file separately and all the tests passed. There is something else occurring. I will keep investigating why this is happening.

rodrigomoliveira1:
> The problem is that matplotlib should raise this error for a polar plot, which is not the case here.
I tried to run the test_fluid_flow.py file separately and all the tests passed. There is something else occurring. I will keep investigating why this is happening.

Same here, I tested test_fluid_flow.py via anaconda prompt and everything was passing

raphaeltimbo:
Fixed the problem. The commit got a little polluted because of the formatting, but basically the problem was the calling the function without providing a new matplotlib axes. This would trigger the `plt.gca()` which would get the current axes that in this case had a polar projection.

Issue #465, opened 2020-03-06
Test/check linting/formatting in CI


nnadeau:
`black` and other formatting and linting tools should be used in appveyor and travis to check that PRs conform to the guidelines

cc https://github.com/openjournals/joss-reviews/issues/2120

raphaeltimbo:
We have added these tools to our CI.

PR #468, opened 2020-03-06
Including black and isort on Travis-CI as a test.

GabrielBachmann:
This PR includes black and isort as a Travis-CI test in a way that the build will fail if the code does not follow black's and isort's guidelines.


GabrielBachmann:
This PR solves the issue #465

ross/fluid_flow/fluid_flow_graphics.py
```diff
@@ -2,7 +2,6 @@
 import numpy as np
 from bokeh.plotting import figure
 from matplotlib import cm
-from mpl_toolkits.mplot3d import axes3d
```
rodrigomoliveira1: Although it seems this import is not used, we cannot remove this line, once we plotting 3D with matplotlib, we need the axes3d. In `results.py`, it happens the same.

rodrigomoliveira1: Here, we could do something like:
```
# fmt: off
from mpl_toolkits.mplot3d import axes3d
# fmt: on
```

GabrielBachmann: This import was removed with Pycharm's Optimize imports.
I think this can be solved by just adding the import!

rodrigomoliveira1: > This import was removed with Pycharm's Optimize imports.
I think this can be solved by just adding the import!

It's good to go now!

rodrigomoliveira1:
@GabrielBachmann there's some kind of conflict again. Maybe related to the PR about removing coverage from the "REQUIRED" in `setup.py`.

PR #478, opened 2020-03-20
Stochastic Ross module - new folder

rodrigomoliveira1:
This PR brings the base structure to Stochastic Ross.

The files follows the hierarchy than ROSS: different files to build the elements and a main file to build the rotors.

The main difference between these new files and and the element files from ROSS is that the stochastic elements files build list of elements instead a single one.

The analysis will also follow the ROSS structure, but it's not added yet.


rodrigomoliveira1:
We have the very basic for the stochastic module right now.
We can create elements and materials with random parameters, random rotors and perform stochastic analysis, such as:
- Stochastic Cambpell Diagram
- Stochastic Frequency Response
- Stochastic Unbalance Response
- Stochastic Time Response

PR #506, opened 2020-03-30
Class MagneticBearingElement was created

davidjgm:
This class creates a magnetic bearing element. 
The element can be created by:
- defining directly the stifness and damping coefficients with the method **__init__**;
- defining the magnetic parameters g0, i0, ag, nw, alpha and the porportional and derivative gains of a PID controller  kp_pid, kd_pid with the method **param_to_coef**.

The reference for the equations is:
**Book**: Magnetic Bearings. Theory, Design, and Application to Rotating Machinery
**Authors**: Gerhard Schweitzer and Eric H. Maslen
**Page**: 84-95 (magnetic parameters) and 354 (PID gains)

### DIFF
--- a/ross/bearing_seal_element.py
+++ b/ross/bearing_seal_element.py
@@ -1322,7 +1322,7 @@ class MagneticBearingElement(BearingElement):
 
     @classmethod
     def param_to_coef(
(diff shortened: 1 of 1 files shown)


rodrigomoliveira1:
@davidjgm you need to run black and isort with your code.

Issue #511, opened 2020-04-02
Bug in the interaction between the Material class and check_units


rodrigomoliveira1:
While writting tests for the new stochastic module, I found a bug instantiating a material.

Example:
- When I write the following code, it runs fine:
```
>>> from ross.materials import Material
>>> mat = Material(name="test", rho=7000, E=None, G_s=81.2e9, Poisson=0.27)
>>> mat
Material(name="test", rho=7.00000e+03, G_s=8.22835e+10, E=2.09000e+11, color='#525252')
```
- But if I ommit the arg names, it returns an error with `check_units`:
```
>>> from ross.materials import Material
>>> mat = Material('test', 7000, None, 81.2e9, 0.27)
Traceback (most recent call last):

  File "D:\Google Drive\Backup notebook\Lavi\ross-master-02-2020\ross\units.py", line 66, in inner
    try:

AttributeError: 'NoneType' object has no attribute 'to'


During handling of the above exception, another exception occurred:

Traceback (most recent call last):

  File "<ipython-input-131-54284e3c45e6>", line 1, in <module>
    mat = Material('test', 7000, None, 81.2e9, 0.27)

  File "D:\Google Drive\Backup notebook\Lavi\ross-master-02-2020\ross\units.py", line 68, in inner
    except AttributeError:

  File "D:\Anaconda3\lib\site-packages\pint\quantity.py", line 194, in __new__
    value, inst.force_ndarray, inst.force_ndarray_like

  File "D:\Anaconda3\lib\site-packages\pint\compat.py", line 71, in _to_magnitude
    raise TypeError("Invalid magnitude for Quantity: {0!r}".format(value))

TypeError: Invalid magnitude for Quantity: None
```
It only happens when passing `E` or `G_s` as None, because both have units related to them.

PR #513, opened 2020-04-03
Fix issue with check_units

rodrigomoliveira1:
Related to Issue #511 

Add a condition to verify if the argument value is None. It allows the user to pass "None" args without getting an error with the code trying to assign a unit to a None value.

### DIFF
--- a/ross/units.py
+++ b/ross/units.py
@@ -59,19 +59,17 @@ def check_units(func):
         args_names = inspect.getfullargspec(func)[0]
 
         for arg_name, arg_value in zip(args_names, args):
-            if arg_name in units:
+            if arg_name in units and arg_value is not None:
                 # For now, we only return the magnitude for the converted Quantity
                 # If pint is fully adopted by ross in the future, and we have all Quantities
                 # using it, we could remove this, which would allows us to use pint in its full capability
-                if arg_value is None:
-                    pass
-                else:
-                    try:
-                        base_unit_args.append(arg_value.to(units[arg_name]).m)
-                    except AttributeError:
-                        base_unit_args.append(Q_(arg_value, units[arg_name]).m)
+                try:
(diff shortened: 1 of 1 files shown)


ross/units.py
```diff
@@ -63,10 +63,13 @@ def inner(*args, **kwargs):
                 # For now, we only return the magnitude for the converted Quantity
                 # If pint is fully adopted by ross in the future, and we have all Quantities
                 # using it, we could remove this, which would allows us to use pint in its full capability
-                try:
-                    base_unit_args.append(arg_value.to(units[arg_name]).m)
-                except AttributeError:
-                    base_unit_args.append(Q_(arg_value, units[arg_name]).m)
+                if arg_value is None:
+                    pass
```
raphaeltimbo, line 67: You still need to add the `arg_value` to the `base_unit_args` list to keep the consistency (args that goes into `check_units` needs to come back).

The code would not pass the following test:
```python
def test_positional_arguments():
    mat = Material('material', 7000, None, 81.2e9, 0.27)
    assert mat.rho == 7000
    assert mat.E == 206248000000.0
    assert mat.G_s == 81.2e9
    assert mat.Poisson == 0.27
```

I recommend adding a check for `arg_value` on line 62:
```python
        for arg_name, arg_value in zip(args_names, args):
            if arg_name in units and arg_value is not None:
                # For now, we only return the magnitude for the converted Quantity
                # If pint is fully adopted by ross in the future, and we have all Quantities
                # using it, we could remove this, which would allows us to use pint in its full capability
                try:
                    base_unit_args.append(arg_value.to(units[arg_name]).m)
                except AttributeError:
                    base_unit_args.append(Q_(arg_value, units[arg_name]).m)
            else:
                base_unit_args.append(arg_value)
```

rodrigomoliveira1: yes, I just noticed the same problem here. Gonna fix it again

PR #531, opened 2020-04-20
6dof assembly

verg1lio:
Generalization of the assembly and evaluation codes of ROSS for a generic "N" DoF sized formulation. This is intended to enable the 6 DoF implementation to be executed appropriately, and is made in a generalized fashion by default.


ross/bearing_seal_element.py
```diff
@@ -1736,17 +1739,17 @@ def load(file_name=""):
...
         if self.n_link is not None:
             global_index = global_index._asdict()
-            global_index[f"x_{self.n_link}"] = 6 * self.n_link
-            global_index[f"y_{self.n_link}"] = 6 * self.n_link + 1
-            global_index[f"z_{self.n_link}"] = 6 * self.n_link + 2
+            global_index[f"u_{self.n_link}"] = 6 * self.n_link
+            global_index[f"v_{self.n_link}"] = 6 * self.n_link + 1
+            global_index[f"w_{self.n_link}"] = 6 * self.n_link + 2
             dof_tuple = namedtuple("GlobalIndex", global_index)
             global_index = dof_tuple(**global_index)
```
rodrigomoliveira1, line 1754: I believe we can remove this method, since the global index are calculated in `Rotor` class.
Instead, add the attribute `self.dof_global_index = None`.

ross/rotor_assembly.py
```diff
@@ -3115,7 +3150,7 @@ def __init__(
             dof_tuple = namedtuple("GlobalIndex", global_dof_mapping)
             elm.dof_global_index = dof_tuple(**global_dof_mapping)
             df.at[
-                df.loc[df.tag == elm.tag].index[0], "dof_global_index"
+                df.loc[df.tag == elm.tag].index[0], "dof_globaxl_index"
```
rodrigomoliveira1, line 3153: Typo here `dof_globaxl_index`, I think.
Make it `dof_global_index`

ross/rotor_assembly.py
```diff
@@ -3452,3 +3487,65 @@ def MAC_modes(U, V, n=None, plot=True):
...
+    Returns
+    -------
+    An instance of a 6DoFs rotor object.
+
+    Examples
+    --------
+    >>> rotor = rs.rotor_assembly.rotor_example_6dof()
+    >>> modal = rotor.run_modal(speed=0)
+    >>> np.round(modal.wd[:4])
+    array([ X  X  X  X ])
```
rodrigomoliveira1, line 3510: I think u meant:

```
>>> import ross as rs
>>> rotor = rs.rotor_example_6dof()
>>> modal = rotor.run_modal(speed=0)
>>> np.round(modal.wd[:4])
array([ X  X  X  X ]) # here, you have to collect the right output
```

ross/rotor_assembly.py
```diff
@@ -198,7 +198,7 @@ def flatten(l):
                 disk.tag = "Disk " + str(i)
 
         for i, brg in enumerate(bearing_elements):
-            if brg.__class__.__name__ == "BearingElement" and brg.tag is None:
+            if isinstance(brg, BearingElement) and brg.tag is None:
                 brg.tag = "Bearing " + str(i)
             if brg.__class__.__name__ == "SealElement" and brg.tag is None:
```
rodrigomoliveira1, line 203: A bearing instance can be a BearingElement and a SealElement at the same time. To avoid this kind of problem, we have to make something like:

```
if not isinstance(brg, SealElement) and brg.tag is None:
    brg.tag = "Bearing " + str(i)
elif isinstance(brg, SealElement) and brg.tag is None:
    brg.tag = "Seal " + str(i)
```

ross/rotor_assembly.py
```diff
@@ -512,6 +518,35 @@ def flatten(l):
...
+
+        for el in self.bearing_elements:
+            if (len(el.dof_mapping()) == 2 and number_dof == 6) or (
+                len(el.dof_mapping()) == 3 and number_dof == 4
+            ):
+                raise Exception(
+                    "The number of degrees o freedom of all elements must be the same! There are BEARING elements with discrepant DoFs."
+                )
+
+        return int(number_dof)
```
rodrigomoliveira1, line 548: We could do it shorter using python's `any()` method.
Please, correct me if there is a problem with the logic.

```
    def _check_number_dof(self):
        """Verify the consistency of degrees of freedom.

        This method loops for all the elements, checking if the number of degrees of
        freedom is consistent.
        E.g.: inputting 2 shaft elements, one with 4 dof and one with 6, will raise
        an error.

        Raises
        ------
        Exception
            Error pointing out difference between the number of DoF's from each element
            type.

        Returns
        -------
        number_dof : int
            Number of degrees of freedom from the adopted shaft element.
        """
        number_dof = len(self.shaft_elements[0].dof_mapping())

        if any(len(sh.dof_mapping()) != number_dof for sh in self.shaft_elements):
            raise Exception(
                "The number of degrees o freedom of all elements must be the same! There are SHAFT elements with discrepant DoFs."
            )

        if any(len(disk.dof_mapping()) != number_dof / 2 for disk in self.disk_elements):
            raise Exception(
                "The number of degrees o freedom of all elements must be the same! There are DISK elements with discrepant DoFs."
            )

        if any(len(brg.dof_mapping()) != number_dof / 4 for brg in self.bearing_elements):
            raise Exception(
                "The number of degrees o freedom of all elements must be the same! There are BEARING elements with discrepant DoFs."
            )

        return int(number_dof / 2)
```

ross/bearing_seal_element.py
verg1lio: Hello Rodrigo, good morning.

We did try to make it by inheriting this method, but due to the different naming it failed. Since it is only a "repeated" code with some modifications, I believe that it is not detrimental to the program.

ross/rotor_assembly.py
verg1lio: Yup, we missed that one, thanks!

ross/rotor_assembly.py
verg1lio: Correct, we will be creating the test functions and proper example values until next Monday, if it all goes according to plan. By then we will have placed proper numerical values for this test.

ross/rotor_assembly.py
verg1lio: Thanks for the tip, we did try some other methods to check the DoF number of each one, but ended up missing this detail.

ross/rotor_assembly.py
verg1lio: Nice! I did not know about the "any()" method. There is a small correction to be made though, the number_dof would be *2, itself and /2, instead of itself, /2 and /4. Other than that i think it would work. We will be analyzing this code and implementing it, thanks again!

ross/bearing_seal_element.py
rodrigomoliveira1: Hello, Vergílio!

You can delete both actually. the `.dof_global_index()` method from 6dof and from 4dof. They're not being called anymore on `Rotor` class. I think we forgot to delete it when we changed the way we build the global indexes.

ross/rotor_assembly.py
rodrigomoliveira1:
```
number_dof = len(self.shaft_elements[0].dof_mapping())
```
number_dof would return 8 or 12 in this code line. The `dof_mapping()` for shaft elements returns de DoF's for 2 nodes. So, its value ends up being doubled.

```len(disk.dof_mapping())``` returns 4 or 6 and ```len(brg.dof_mapping())``` returns 2 or 3. We compare the number_dof, dividing itself by 2 or 4, or multiplying the `len` by 2 or 4.

Comment (rodrigomoliveira1) [reply ross/rotor_assembly.py:548]: Oh, I missed what you've done before. On your code, you already divided the `number_dof` by 2 in `number_dof = len(self.shaft_elements[0].dof_mapping()) / 2`. Sorry.

File: ross/bearing_seal_element.py (hunk 1 above)
Comment (verg1lio) [reply ross/bearing_seal_element.py:1754]: Oh, alright then. We will check and delete it.

File: ross/rotor_assembly.py [hunk 6]
```diff
@@ -519,31 +519,43 @@ def flatten(l):
...
-                raise Exception(
-                    "The number of degrees o freedom of all elements must be the same! There are BEARING elements with discrepant DoFs."
-                )
+        Returns
+        -------
+        number_dof : int
+            Number of degrees of freedom from the adopted shaft element.
+        """
+
+        number_dof = len(self.shaft_elements[0].dof_mapping())
```
rodrigomoliveira1: @verg1lio I think you need to divide by 2 in the end of the line
`len(self.shaft_elements[0].dof_mapping())` returns 8 or 12 (related to 2 nodes). And we need the number of DoF's for a single node, in this case. Thus, the other comparations will make sense.

verg1lio: Hi @rodrigomoliveira1 , good morning. We realized that and did it on the last commit. It seems you have an outdated version there.

ross/rotor_assembly.py
```diff
@@ -3504,10 +3516,16 @@ def rotor_example_6dof():
...
+    >>> import ross as rs
+    >>> import numpy as np
+    >>> import matplotlib.pyplot as plt
+    >>> rotor6 = rs.rotor_assembly.rotor_example_6dof()
+    >>> camp6 = rotor6.run_campbell(np.linspace(0,400,101),frequencies=18)
+    >>> camp6.plot(plot_type="matplotlib")
+    >>> plt.show()
+    >>> modal6 = rotor6.run_modal(speed=0)
+    >>> np.round(modal6.wd[:6])
+    array([  0.,   0.,  48.,  92.,  96., 275.])
```
rodrigomoliveira1: This example will also return the plot called on `camp6.plot(plot_type="matplotlib")`, so, the CI's will claim it and return an error. You can make something like this:
```
    >>> import ross as rs
    >>> import numpy as np
    >>> rotor6 = rs.rotor_example_6dof()
    >>> camp6 = rotor6.run_campbell(np.linspace(0,400,101), frequencies=18)
    >>> camp6.plot(plot_type="matplotlib") # doctest: +ELLIPSIS
    (<Figure...
    >>> modal6 = rotor6.run_modal(speed=0)
    >>> np.round(modal6.wd[:6])
    array([  0.,   0.,  48.,  92.,  96., 275.])
```

verg1lio: Oh alright, I did not know about that one. we will change it in the code right away.

ross/rotor_assembly.py
```diff
@@ -3524,8 +3524,8 @@ def rotor_example_6dof():
     >>> camp6.plot(plot_type="matplotlib")
     >>> plt.show()
     >>> modal6 = rotor6.run_modal(speed=0)
-    >>> np.round(modal6.wd[:6])
-    array([  0.,   0.,  48.,  92.,  96., 275.])
+    >>> modal6.wd[:6]
+    array([  0.        ,  47.62157215,  91.79835717,  96.29386819,  274.51274397, 296.49903736])
```
rodrigomoliveira1, line 3528: Hello @verg1lio!

I was checking for errors reported by APPVEYOR and this example has something wrong. You've changed the output of `modal6.wd[:6]` and now we're getting an error. It seems that the output you set before was correct, look:

image

ross/rotor_assembly.py
```diff
@@ -1744,7 +1791,7 @@ def plot_ucs(self, stiffness_range=None, num=20, ax=None, output_html=False):
                 self.shaft_elements, self.disk_elements, bearings, n_eigen=16
             )
             modal = rotor.run_modal(speed=0)
-            rotor_wn[:, i] = modal.wn[:8:2]
+            rotor_wn[:, i] = modal.wn[: self.number_dof * 2 : self.number_dof / 2]
```
rodrigomoliveira1, line 1794: image

Here, we also have an error message. It says list index must be an interger of None. `self.number_dof` is a int by definition, however, when you divide it by 2, it becomes a float. So, you can solve it doing:

`rotor_wn[:, i] = modal.wn[: self.number_dof * 2 : int(self.number_dof / 2)]`

or similar

verg1lio: Oh, we have had this problem with the 'number_dofs' becoming float everywhere... thanks for noticing this one! I will correct it right away.

rodrigomoliveira1: You're welcome!

Ahh, instead of adding a `int()` command, like a I said, do:
`rotor_wn[:, i] = modal.wn[: self.number_dof * 2 : self.number_dof // 2]`
it will force an integer output.

ross/rotor_assembly.py
verg1lio: Hello @rodrigomoliveira1 , good morning! We had the result with 2 rigid body modes when utilizing the axial stiffness of the bearing equal zero. But when we applied a reasonable value it does change to having only one rigid body mode. I suppose that we are getting still 2 rigid body modes because of an error in the entry data for the example. I will be looking into it now, and will try to fix the issue.

ross/rotor_assembly.py
verg1lio: Done

ross/rotor_assembly.py
verg1lio: I re-ran the same example manually, in a python terminal here, and got the expected results without errors. Also tested the test function examples, getting a pass from the test explorer. Maybe the APPVEYOR is using the null value for the axial bearing stiffness?

rodrigomoliveira1: I don't know. 

Try checking if the code you just run is the same you uploaded to this PR.
Or, try downloading your own branch from github and running it on another instance. It have already happened to me, getting different results from the local code and the repository code. I changed something and forgot to commit it.

raphaeltimbo: I agree with @rodrigomoliveira1 , maybe you have something different in your local code @verg1lio .
I have created a branch from your PR and I am getting two errors in the `rotor_assembly.py` docstrings.
I have also tried to check Travis CI, but it looks like we are getting errors from isort. I have created an issue #538 to discuss if we should remove isort from our CI.

verg1lio: Alright, we will look into this to see where is the problem coming from.

ross/bearing_seal_element.py
```diff
@@ -1736,17 +1739,17 @@ def load(file_name=""):
         return bearing_elements
 
     def dof_mapping(self):
-        return dict(x_0=0, y_0=1, z_0=2)
+        return dict(u_0=0, v_0=1, w_0=2)
```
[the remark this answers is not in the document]
verg1lio: Hello @raphaeltimbo , we kept the standard used in the definition of the element matrices. The differences are cosmetic, so if there is the need, we can change back to (x,y,z) without altering the code function.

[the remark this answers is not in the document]
raphaeltimbo: I agree these don't alter the code function. I am more worried about the documentation for the user. There are parts of our documentation using `x`, `y`... So I think it would be better to follow this convention so that someone reading the docs don't get confused.

ross/rotor_assembly.py
verg1lio: Ok... we detected the problem. 

When executing the tests individually, it passes. When running all the tests from "test_rotor_assembly" together, it fails with your error message. 

Have you ever encountered this issue before?

rodrigomoliveira1: Check if you have duplicated ROSS folder salved on another diretory.

When you run it individually, the code refers to the files you just opened and you're using right now. When running all the test at once, maybe, it refers to another ROSS folder.

And check if, in the test files, the imports have `ross.`

verg1lio: We are seeing the same problem on different machines. So, we will check the first 6 non-zero terms of the natural frequencies, to avoid any problem issuing from the "eig" solution on different machines. It should solve the problem. 

I will get back to you here once that is solved.

verg1lio: We noticed the same problem related to the "eig" solution on different machines and different code runs, due to the "eig" coding. There are a few comments like the following:

@pytest.mark.skip(reason="Different evector order when not sorted")

and the problem we are facing is most likely the same.

ross/bearing_seal_element.py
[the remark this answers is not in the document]
verg1lio: Corrected here (https://github.com/ross-rotordynamics/ross/pull/531/commits/1c3dc0992e4d5833d3e3f5a9fc3f3b693e39cec4)

rodrigomoliveira1:
@verg1lio nice work! I'll just add a commit to your branch fixing the isort issue, that TRAVIS is blocking, ok?

raphaeltimbo:
Hi there, @verg1lio ! Is this still a work in progress or can we merge it?

verg1lio:
Hi @raphaeltimbo , good morning!

Responding you on (https://github.com/ross-rotordynamics/ross/pull/531#issuecomment-621149947), I believe that this part  of the rotor system assembly and Modal+Campbell analyses is completed. There is still the issue with numerical precision on the eigenvalue solution, but that is another problem altogether. All the tests we ran, on different machines, indicate that, and also indicate that our code runs as intrended.

Prof. @rodrigomoliveira1 already checked the implementation, and if it is ok for him i suppose it is fine to merge.

rodrigomoliveira1:
Hello @verg1lio! Ok, so I'm removing the `WIP` tag and merging the PR.

rodrigomoliveira1:
@raphaeltimbo Do you have any other change requests?

verg1lio:
Good morning! Following up on what Prof. Aldemir informed me, we will be continuing the development and generalization of the ROSS code. The goal is to finish generalizing the analyses for the 6 DoF model.

PR #540, opened 2020-05-02
Sparse matrix method

PedroBernardino:
Change in the method used to solve the linear system in the resolves_matrix method of the FluidFlow Class

### DIFF
--- a/ross/fluid_flow/fluid_flow.py
+++ b/ross/fluid_flow/fluid_flow.py
@@ -4,6 +4,8 @@ import sys
 
 import numpy as np
 
+import scipy as sp
+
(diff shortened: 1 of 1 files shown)


raphaeltimbo:
Hi there, @PedroBernardino ! I saw that this has already been merged, but I have a comment so if you could adress this latter.

It is better to use:
```python
from scipy import sparse
```
instead of `import scipy as sp`.
In principle there is nothing wrong with import scipy as sp, but it is recommended to follow these guidelines:
https://docs.scipy.org/doc/scipy-0.18.1/reference/api.html#guidelines-for-importing-functions-from-scipy
so we have a more standard and easy to read code.

PR #570, opened 2020-05-28
Add deflected shape plot method

rodrigomoliveira1:
This PR has only a debugging purpose for the deflected shape plot method.
We are checking methods to calculate the major axis for the nodal orbits.


rodrigomoliveira1:
@raphaeltimbo I'll open a new PR related to this topic later. I'm almost done with the modifications.

Issue #569, opened 2020-05-27
Deflected shape diagram to unbalance response


rodrigomoliveira1:
@atbrandao, following up on your previous idea about improving the forced response (Issue #553). Would you be interested on a 3D deflected shape diagram for the rotor?
The diagram would display the calculated nodal orbits for a given `frequency`. The unbalances would be displayed as arrows on the nodes where they're applied, also indicating its phase.

The `frequency` parameter could be optional, The user could choose any frequency within the `frequency_range` argument passed to the method. The default'd be the 1st critical frequency.

@raphaeltimbo I'd like to hear your opinion too.

atbrandao:
@rodrigomoliveira1 , this would be incredibly useful! And since you came up with this idea, I'd like do go further and expand it:

- Also provide, the 2D deflected shapes showing the nodal major axis and minor axis. Some times, for complex modes, it is easier to have a physical idea of the mode shape when it is in 2D, even if it neglects the relative phases between nodes. 

- Provide a Bending Moment vs Shaft Length plot for the chosen frequency. I had to evaluate this in a case where we had severe misalignment-like forces and I was concerned about the stresses in the shaft. I had to do it on Matlab, because none of the Rotordynamic tools we use have this kind of output. A second step would be to consider surface finish and fillet radii to output the actual stresses acting on the shaft. **Maybe this should become a separate Issue....**

rodrigomoliveira1:
Good to hear that! So, I'm going to work on this feature.

> Also provide, the 2D deflected shapes showing the nodal major axis and minor axis. Some times, for complex modes, it is easier to have a physical idea of the mode shape when it is in 2D, even if it neglects the relative phases between nodes

I'll add the option for plotting 2D either.

> Provide a Bending Moment vs Shaft Length plot for the chosen frequency. I had to evaluate this in a case where we had severe misalignment-like forces and I was concerned about the stresses in the shaft. I had to do it on Matlab, because none of the Rotordynamic tools we use have this kind of output. 

Good to know. And it's not difficult to implement this diagram. 
I'm only concerned about the code organization. For example, now we have a `.plot()` method that calls 3 other sub-methods (`plot_magnitude()`, `plot_phase()`. and `plot_polar()`) and groups the graphs like I showed in Issue #553. Adding more plots to that grid would cause the graphs to become too small and there would be a lot of information at the same time. I'll think about a solution to it, maybe I create a `plot_deflected_shape()` method to display these new graphs.

[...]

PR #574, opened 2020-06-01
Add methods to plot the deflected shape

rodrigomoliveira1:
Related to Issue #569.

After running the method `.run_unbalance_response()` the user is able to call a `.plot()` method to display the amplitude and phase vs frequency response.
This PR adds an alternative method to plot a 3D orbit (rotor deflected shape) response for a given unbalance: `.plot_deflected_shape()`.

```
>>> rotor = rotor_example()
>>> speed_range = np.linspace(0, 1000, 101)
>>> response = rotor.run_unbalance_response(node=3,
...                                         magnitude=10.0,
...                                         phase=0.0,
...                                         frequency_range=speed_range)
# plot unbalance response:
>>> fig = response.plot(dof=13)
# plot deflected shape:
>>> speed = 900
>>> fig = response.plot_deflected_shape(speed)
```

The image below is the current output:

image

### DIFF
--- a/ross/results.py
+++ b/ross/results.py
@@ -1273,13 +1273,19 @@ class ForcedResponseResults:
     Parameters
     ----------
     force_resp : array
-        Array with the force response for each node for each frequency
+        Array with the force response for each node for each frequency.
     speed_range : array
-        Array with the frequencies
+        Array with the frequencies.
     magnitude : array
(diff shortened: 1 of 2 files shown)


rodrigomoliveira1:
Notice that there's still an empty space. There'll be located the bending moment diagram, which I did not finish yet.

Issue #580, opened 2020-06-08
Plot style


raphaeltimbo:
Since we are changing all plots to plotly, I wanted to use this opportunity to discuss how we will change the style for each plot.

One point that I have in mind is that we should not change the plots too much from the plotly default style. This will save us energy on having to write/maintain fewer lines of code and also on keeping the consistency between the styles of different plots (fonts, colors, grids etc.).

Another point is that plotly has a theme/template philosophy that is similar to that of matplotlib, so if the user needs to change the style of multiple plots for a different context (e.g. from an article to a presentation) they could do that by changing the style. As an example, when we change the theme to 'presentation' this will increase the font size for the labels to make it easier to visualize. This will not work if we change the font sizes in our plots.

Plotly has a list of themes/templates here: https://plotly.com/python/templates/ 
My suggestion is that we choose one of these themes and make the modifications to it in the top of the module, and them we should keep modifications on each plot to a minimum.

rodrigomoliveira1:
Good idea. Although I like the available templates on Plotly, we could create our own.
I've "copied and pasted" the layout parameters for all the plots, and indeed it's too much repetitive, and one can easily skip changing one of the plot styles when making modifications.

We could standarize:
- Plot Layouts:
    - title font (family, size)
    - plot background color
    - hoverlabel alignment
    - legend
        - font (family, size)
        - background color,
        - border color,
        - border width
- Axes Layout:
    - title font (family, size),
    - tick font (family, size),
     - grid color,
     - showline,
     - line width,
     - linecolor,
     - mirror,

Maybe we could also do it to `coloraxis_colorscale`, when creating ColorBars, but I don't know if it's gonna work well for Scatter colors.

raphaeltimbo:
I will start to work on a script to generate all plots for 2 or 3 different models. 
This way we will be able to evaluate how each modification affects different plots.

PR #582, opened 2020-06-11
Add plotly theme

raphaeltimbo:
This PR is related to #580.
We want to create a plotly theme to avoid having to set every style parameter for each plot, saving us lots of lines of code, decreasing burden in maintaining the code and keeping the style consistent across different plots.


ross/plotly_theme.py
```diff
@@ -0,0 +1,377 @@
...
+    "orange": "#ff7f0e",
+    "green": "#2ca02c",
+    "red": "#d62728",
+    "purple": "#9467bd",
+    "brown": "#8c564b",
+    "pink": "#e377c2",
+    "gray": "#7f7f7f",
+    "olive": "#bcbd22",
+    "cyan": "#17becf",
+}
```
rodrigomoliveira1, line 16: Wouldn't it be simpler to import this palette directly from plotly?
I mean:
```
from plotly.express import colors
tableau_colors = colors.qualitative.T10
```

ross/plotly_theme.py
```diff
@@ -0,0 +1,377 @@
...
+                [0.3333333333333333, "#9c179e"],
+                [0.4444444444444444, "#bd3786"],
+                [0.5555555555555556, "#d8576b"],
+                [0.6666666666666666, "#ed7953"],
+                [0.7777777777777778, "#fb9f3a"],
+                [0.8888888888888888, "#fdca26"],
+                [1.0, "#f0f921"],
+            ],
+        },
+        "colorway": list(colors_dict.values()),
```
rodrigomoliveira1, line 64: And here, it would change to `colorway = tableau_colors`

ross/plotly_theme.py
```diff
@@ -0,0 +1,377 @@
...
+                    [0.1111111111111111, "#46039f"],
+                    [0.2222222222222222, "#7201a8"],
+                    [0.3333333333333333, "#9c179e"],
+                    [0.4444444444444444, "#bd3786"],
+                    [0.5555555555555556, "#d8576b"],
+                    [0.6666666666666666, "#ed7953"],
+                    [0.7777777777777778, "#fb9f3a"],
+                    [0.8888888888888888, "#fdca26"],
+                    [1.0, "#f0f921"],
+                ],
```
rodrigomoliveira1, line 203: Here, it's the same idea. This color scale is blue to red (I suppose by testing the values), and we could replace the list by:
`"rdbu"` or `colors.sequential.RdBu`

ross/plotly_theme.py
raphaeltimbo: We could do that, but the problem is that `colors.qualitative.T10` returns just a list with hex values and I wanted to have a dictionary so that in other parts of the code we could use colors_dict['blue'] instead of `tableau_colors[0]`.

ross/plotly_theme.py
raphaeltimbo: This one is just a copy from the plotly theme. To me it is ok if we replace it.
I will make a document with all plots that we have (ucs, campbell, frequency response etc.) so that we can see how the modifications to the plotly theme affects all figures and then I will test the modification and show it here.

rodrigomoliveira1: Got it, ok then

ross/results.py
```diff
@@ -1072,39 +1000,14 @@ def plot_phase(self, inp, out, **phase_kwargs):
...
-        )
-        fig.update_layout(
-            legend=dict(
-                font=dict(family="sans-serif", size=14),
-                bgcolor="white",
-                bordercolor="black",
-                borderwidth=2,
-            ),
-            **phase_kwargs,
+            title_text="<b>Frequency</b>", range=[0, np.max(frequency_range)]
```
rodrigomoliveira1, line 1003: I think we have to fix this range. I did a bad default range, setting to begin at 0. If the user does not input a frequency_range starting at 0, the plot x axis will still start at 0.
I fixed it in run_campbell, but I let it pass in other classes.

Change
`range=[0, np.max(frequency_range)]`
to 
`range=[np.min(frequency_range), np.max(frequency_range)]`
or similar

ross/results.py
```diff
@@ -1372,46 +1253,22 @@ def plot_magnitude(self, dof, units="m", **kwargs):
...
-        )
-        fig.update_layout(
-            legend=dict(
-                font=dict(family="sans-serif", size=14),
-                bgcolor="white",
-                bordercolor="black",
-                borderwidth=2,
-            ),
-            **kwargs,
+            title_text="<b>Frequency</b>", range=[0, np.max(frequency_range)]
```
rodrigomoliveira1, line 1256: Same here.

ross/results.py
```diff
@@ -1447,39 +1299,14 @@ def plot_phase(self, dof, **kwargs):
...
```
rodrigomoliveira1, line 1302: Same here.

ross/results.py
raphaeltimbo: I will do this now.

raphaeltimbo:
So, I think this PR has too many modifications already. 
Maybe some plots have lost some formatting that I should not have removed, but I think we can solve this in a different PR. 
The purpose for now was just to remove all formatting and try to move everything to the plotly_theme. I have also added fig as an argument to all plots.

Issue #607, opened 2020-06-30
Weasyprint not easily installing in windows


GabrielBachmann:
To generate a report as a pdf we are using weasyprint, but this module has shown it's downsides when trying to installing it on windows, there's not a way to do it only with pip.

GabrielBachmann:
We've decided to not use weasyprint, using dash instead.

GabrielBachmann:
Dash does not support the export of an serverless HTML file so we also need to have a Jinja2 template.

PR #606, opened 2020-06-30
WIP - Change report to dash format and restructure of Report folder

GabrielBachmann:
Formely, we were using weasyprint to generate a pdf static report.
There was some installation problems with weasyprint on windows, so we decided to change weasyprint for Dash, besides that, dash gives an dynamic html report and the option to download as a pdf.

PR #613, opened 2020-07-02
WIP - Changing report format to use dash and jinja2 

GabrielBachmann:
As discussed in #606 and #607, we've decided to change weasyprint for jinja2 and dash.

There's 4 main methods used in the html generation.
Those are: 
* run();
* assets_prep();
* generate_report();
* export_html();

The first one runs all analysis that is placed on the report and returns a dictionary with divided plotly Figures and pandas DataFrames.

The second one prepare the assets to be put in the reports and it returns a dict contaning all plotly figures and the HTML version of them.

The third one generates a dash version of the report.

The last one exports an html file, it uses jinja2 and exported plotly Figures as HTML, the only reason we have this function is because dash does not support a "serverless" option to share dashboards.


rodrigomoliveira1:
@GabrielBachmann, can we close this PR and reopen it in ROSS_Report repository? I'm willing to remove these files.

GabrielBachmann:
This PR is now closed and all the contents in it will be moved to another [repo](https://github.com/ross-rotordynamics/ross-report).

PR #626, opened 2020-07-10
Add units to plot functions

### LINKED ISSUE #609: Units support
ROSS has only some small support for units implemented with the `check_units` decorator.
This decorator will verify each argument name and convert to units declared in the units.py module and then it will return the converted values to the function. 
The returned values are of the same type as the original argument (float or int), and are not of the pint.Quantity type. This is because after converting with pint.Quantity we only pass the magnitude attribute.
Ideally we would have pint.Quantity objects passed and everything would have a magnitude/unit attached to it, so when creating a material the user could run things like:
```python
import ross as rs

Q = rs.units.Q
steel = rs.Material(name='steel', rho=7810, E=Q(3e7, 'psi'), G_s=Q(1e7, 'psi'))

print(steel.E)
206842721646.701 N / m2
print(steel.E.to('lbf/inch**2'))
30000000.0 lbf / in2
print(steel.rho.to('lb/ft**3'))
487.56237209968947 lb / ft3
```

I have tried to implement this approach here https://github.com/raphaeltimbo/ross-1/tree/improve-pint but I had some problems.
1. Working with arrays and units in our case is not trivial. Take, for example, the disk mass matrix:
```python
M = np.array([[m, 0,  0,  0],
              [0, m,  0,  0],
              [0, 0, Id,  0],
              [0, 0,  0, Id]]
```
Here `m` has units kg and `Id` kg*m**2, so it is not possible to assemble a `Quantity` array in this case.
So we will have to use the Quantity magnitude to construct the array:
```python
m = m.magnitude
Id = Id.magnitude
M = np.array([[m, 0,  0,  0],
              [0, m,  0,  0],
              [0, 0, Id,  0],
              [0, 0,  0, Id]]
```
The problem with this is that we kind of defeat the purpose of having Quantity objects passed to these functions if we have to fall back to using their magnitude all the time.
2. Other modules depending on ROSS main modules would have to adapt to expect Quantity objects. For example, the following test for the sthocastic module fails with this code:
```python
def test_st_shaft_element(rand_shaft):
    elm = list(iter(rand_shaft))
    assert [sh.L for sh in elm] == [1.0, 1.1]
    assert [sh.idl for sh in elm] == [0.01, 0.02]
    assert [sh.odl for sh in elm] == [0.1, 0.2]
    assert [sh.idr for sh in elm] == [0.01, 0.02]
    assert [sh.odr for sh in elm] == [0.1, 0.2]
    assert [sh.material.E for sh in elm] == [209000000000.0, 211000000000.0]
```
Result:
```python
Expected :[1.0, 1.1]
Actual   :[<Quantity 1. m>, <Quantity 1.1 m>]
```
Using Quantity everywhere will also add some extra computing time.

So, considering that, I think it is not worth for our package to have pint 'fully' implemented.
We could keep working with the unit conversion in the input and we could also add some utility functions to help the user converting the final results and also plotting with the desired units.
Internally, however, we would keep working with normal arrays/floats.

I will start to implement something and will send updates here for some feedback.

raphaeltimbo:
This PR adds arguments to the plot functions so that the user can control the plot units.
As an example, for the unbalance response, now it is possible to call the plot function like this:
```python
unb = rotor.run_unbalance_response(
    node=23, unbalance_magnitude=0.60458, unbalance_phase=0, frequency=freq_range,
)
unb.plot(3, frequency_units='RPM')
```

This will automatically convert the frequency units from the default 'rad/s' to 'RPM'.

There are cases where we could have one argument controlling all quantities in the plot, like:

```python
def plot_deflected_shape_3d(
    self,
    speed,
    samples=101,
    length_units="m",
    fig=None,
    **kwargs,
):
```
The problem with this is that, if the user wanted to see the displacement in each node in `microns` the rotor length would also be changed to `microns`, and the plot would be strange since the displacement and the rotor length are not in the same order of magnitude.
To solve this, I have added two arguments to give the user more control:

```python
def plot_deflected_shape_3d(
    self,
    speed,
    samples=101,
    displacement_units="m",
    rotor_length_units="m",
    fig=None,
    **kwargs,
):
```

Closes #609.

### DIFF
--- a/ross/bearing_seal_element.py
+++ b/ross/bearing_seal_element.py
@@ -18,7 +18,7 @@ from ross.element import Element
 from ross.fluid_flow import fluid_flow as flow
 from ross.fluid_flow.fluid_flow_coefficients import (
     calculate_short_damping_matrix, calculate_short_stiffness_matrix)
-from ross.units import check_units
+from ross.units import Q_, check_units
 from ross.utils import read_table_file
 
 # fmt: on
@@ -160,11 +160,15 @@ class _Coefficient:
         """
         return self.coefficient[item]
 
-    def plot(self, **kwargs):
+    def plot(self, frequency_units="rad/s", y_units=None, fig=None, **kwargs):
         """Plot coefficient vs frequency.
 
         Parameters
         ----------
+        frequency_units : str
+            Frequency units.
+            Default is rad/s.
+        y_units : str
         **kwargs : optional
             Additional key word arguments can be passed to change the plot layout only
             (e.g. width=1000, height=800, ...).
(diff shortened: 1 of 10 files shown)


rodrigomoliveira1:
There's still a conflict, although you have fixed it on last commit.

raphaeltimbo:
@rodrigomoliveira1 , here it was not showing any conflicts, so I could merge it.

rodrigomoliveira1:
Perhaps it's a bug in GitHub page

PR #635, opened 2020-07-23
Add probe argument to ForcedResponseResults

rodrigomoliveira1:
- Replace `dof` by `probe `argument
- probe is a list of tuples. Each tuple refers to a probe, which must include de node and the orientation.
- The orientation units is set by `probe_units` argument.
- in `plot()` method, `data.showlegend` is turn off to avoid repetitive legend marks on graphs

### DIFF
--- a/ross/api_report.py
+++ b/ross/api_report.py
@@ -780,8 +780,8 @@ class Report:
 
         for node in nodes:
             dof = 4 * node + 1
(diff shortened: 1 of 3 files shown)


rodrigomoliveira1:
@raphaeltimbo, I did a generic modification to api_report.py, only to run the tests, since we're moving this file and it's pretty behind from what we have on ROSS_Report repository.

Issue #647, opened 2020-08-12
Error in the eccentricity ratio for approaching short bearings


JuliaMota:
There is a problem with the function that calculates the eccentricity ratio from the Sommerfeld number for approaching short bearings. The function considers a 4th degree polynomial while it should be 8th degree. In addition, there are improvements that can be made in order to obtain a better approximation to the eccentricity in the equilibrium position.

It is worth remembering that the short bearing is widely used and the balance position found by this method is used as an initial kick for the iterative method that seeks the balance position for bearings of any size.

alannakamura:
To find the correct eccentricity rate, the roots of the fourth degree polynomial were taken, complex roots were eliminated, and the real roots were ordered in ascending order. Then I chose the smallest root that had a value between 0 and 1 and returned its square root. The square root was necessary because the polynomial solved in the code was a fourth degree polynomial, so that the square of the eighth degree root was found.

PR #650, opened 2020-08-13
Correct the eccentricity ratio in the short bearing

JuliaMota:
Corrected the function that calculates the eccentricity ratio in the short bearing.
This pull request closes the issue #647

### DIFF
--- a/ross/fluid_flow/fluid_flow_geometry.py
+++ b/ross/fluid_flow/fluid_flow_geometry.py
@@ -233,16 +233,8 @@ def calculate_eccentricity_ratio(modified_s):
         -(4 + (np.pi ** 2) * (modified_s ** 2)),
         1,
     ]
-    roots = list(np.roots(coefficients))
-    roots2 = roots.copy()
-    for i in roots2:
-        if not np.isreal(i):
-            roots.remove(i)
-    del roots2
-
-    roots.sort()
-    for i in range(len(roots)):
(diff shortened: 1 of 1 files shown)


ross/fluid_flow/fluid_flow_geometry.py
```diff
@@ -232,10 +233,40 @@ def calculate_eccentricity_ratio(modified_s):
...
+    roots = list(np.roots(coefficients))
+    roots2 = roots.copy()
+    for i in roots2:
+        if not np.isreal(i):
+            roots.remove(i)
+    del roots2
+
+    roots.sort()
+    for i in range(len(roots)):
+        roots[i] = roots[i].real
```
rodrigomoliveira1, line 245: `np.isreal`  checks for the imaginary part being exactly zero. Couldn't it be a problem if a root has a very small imaginary part, like `1e-16` due numerical truncation error?
```python
>>> np.isreal([1 + 1e-16j])
array([False])

>>> np.isreal([1 + 0j])
array([True])
```

Considering it's not a problem, you could rewrite this code like:
```python
    roots = np.roots(coefficients)
    roots = np.sort(roots[np.isreal(roots)].real)
```
Working with numpy array is more efficient than with pure lists.

JuliaMota: You're right. This way I can reach the same result.

PR #652, opened 2020-08-17
Implements Misalignment

hiagopinacio:

PR #689, opened 2020-10-09
Fix #686

### LINKED ISSUE #686: Unexpected errors in `generate_comparasion_plot.py`
In PR #652, @verg1lio, @bela-bs and @marcusfilipesr are experiencing some unexpected errors with `generate_comparasion_plot.py` while testing the new defect module.

We could add some routines to check the correct functioning of this file.

@verg1lio, it would be very useful if you could post here what exactly is happening and the log error.

marcusfilipesr:
Correcting problem cited in issue https://github.com/ross-rotordynamics/ross/pull/652#issuecomment-705647685.


marcusfilipesr:
Good afternoon @raphaeltimbo @rodrigomoliveira1, we've created this PR to solve the discussed problem regarding the issue https://github.com/ross-rotordynamics/ross/pull/652#issuecomment-705647685. After we have your approval we will commit the proposed changes.

tools/generate_comparison_plots.py
```diff
@@ -1,3 +1,10 @@
+import os.path
+import sys
+
+sys.path.append(
+    os.path.abspath(os.path.join(os.path.dirname(__file__), os.path.pardir))
```
raphaeltimbo, line 5: Hi there, guys! 
My only comment here is that in the ROSS code base we are trying to have the pathlib library as a standard for taking care of paths. 

Here the code would be something like:

```python
import sys

from pathlib import Path

import numpy as np
import plotly.io as pio
from jinja2 import Environment, FileSystemLoader

sys.path.append(Path(__file__).parent.parent)
import ross as rs
```

raphaeltimbo (changes requested):

raphaeltimbo (changes requested):
Hi there, guys! Thanks for fixing this.
I have only made one comment regarding the use of pathlib.

tools/generate_comparison_plots.py
marcusfilipesr: We'll fix this right away @raphaeltimbo!

marcusfilipesr:
Hi @raphaeltimbo, we've just made the required changes. Just a curiosity for you guys, while we were searching on how to use the `pathlib` library for changing the sys path, we found that the method `sys.path.append()` requires a string parameter. So using your tip, we did have to change only that, as it follows:
```python
import sys
from pathlib import Path

import numpy as np
import plotly.io as pio
from jinja2 import Environment, FileSystemLoader

sys.path.append(str(Path(__file__).parent.parent))
import ross as rs
```
Just in case you guys ran into a similar situation.

rodrigomoliveira1 (approved):

raphaeltimbo (approved):

Issue #687, opened 2020-10-08
Integrating ROSS and ROSS-ML through new AnnBearingElement classes 


rodrigomoliveira1:
As the new ROSS machine learning module (`ROSS-ML`) takes shape, we could add new features to ROSS aiming the integration of both packages.

`ROSS-ML` is about building neural networks aiming (mainly) to predict rotordynamics coefficients for bearings and seal elements. Hence, out integration goes through `bearing_seal_element.py` file. 

### Class structure
The idea is creating new `**BearingElement` classes that allow the user inputs a previously built neural network and a new set or parameters which will be used to predict the stiffness and damping coefficients.
```python
element = NewBearingElementClass(neural_network, **set_of_parameters)
```

Aware of the fact that each neural network model can request different parameters, we need to create a class that is able to receive any type of neural network, regardless of how many and which are the parameters. And, at the same time, we need to keep ROSS standards of instantiating `BearingElement`.

### How to do it?
First, we create a general class `AnnBearingElement` (ANN stands for artificial neural network) that accepts any kind of network model and any parameters (`**kwargs`).
This class loads the model and compares its features to the `**kwargs` passed. If they match, the class procceds to use the neural network to predict the coefficients, if not, the class raises an `KeyError` pointing out the exactly parameters the user should enter (the model features).

At this point, the class documention would not be very clear about what parameters to input, once the user would have to load the model and check what are the features.
```python
>>> import rossml as rsml
>>> model = rsml.Model("my_neural_network")
>>> model.features
["feature1", "feature2", "feature3", ...]
```
Besides, it wouldn't explain what type each feature represents (float, int, string, ...) and what they represent.

Now, let's check how this class is built
```python
class AnnBearingElement(BearingElement):
    """Create bearings or seals elements via Artificial Neural Netwarks.

    This class creates a element from a trained neural network. In order to create it
    properly, one should bear in mind the name of the trained network, and the
    parameters used. The results will be given in a dataframe form. When the number of
    features is different from the trained one, a error is displayed and the same
    occurs when the variable name is wrong due to a typo.

    Parameters
    ----------
    arq : str
        The neural network folder's name, which must be located at ross-ml package.
        The model files are loaded from this folder.
    n : int
        The node in which the element will be located in the rotor.
    n_link : int, optional
        Node to which the bearing will connect. If None the bearing is
        connected to ground.
        Default is None.
    scale_factor : float, optional
        The scale factor is used to scale the bearing drawing.
[...]

PR #698, opened 2020-11-10
New BearingElement classes based on neural networks 

rodrigomoliveira1:
Related to the discussion #687.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,3 +6,4 @@ pandas>=0.23
 plotly>=4.7.0
 xlrd
 pint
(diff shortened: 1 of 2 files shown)


rodrigomoliveira1:
The file called `random_sampler.py` from `ross-ml` is causing this errors. It is not relevant to the package and will be removed. Then, a new release should be deployed to PyPI.

rodrigomoliveira1:
I'm canceling this PR to move these new classes to ROSSML, as it has been spoken in Issue #687.

PR #702, opened 2020-11-23
Crack 

bela-bs:
Implements crack defect


bela-bs:
Good Morning @raphaeltimbo and @rodrigomoliveira1. This PR is for the implementation of codes that simulate crack defects. Any suggestions or questions, you can tell us. 
Izabela Batista

ross/defects/crack.py
```diff
@@ -496,9 +498,9 @@ def _mayes(self, ap):
         return K
 
     def _get_coefs(self, coef):
-        x = scipy.io.loadmat(
-            "/home/izabela/Documents/Projeto EDGE Petro/ross/tools/data/PAPADOPOULOS_c"
-        )
+
+        cwd = os.getcwd()
+        x = scipy.io.loadmat(cwd + "/tools/data/PAPADOPOULOS_c")
```
raphaeltimbo, line 503: I would recommend using pathlib for this.
It is the library that we are using in most of our code for dealing with paths.

ross/defects/crack.py
```diff
@@ -0,0 +1,509 @@
...
+            - np.cos(19 * ap) / 19
+            + np.cos(21 * ap) / 21
+            - np.cos(23 * ap) / 23
+            + np.cos(25 * ap) / 25
+            - np.cos(27 * ap) / 27
+            + np.cos(29 * ap) / 29
+            - np.cos(31 * ap) / 31
+            + np.cos(33 * ap) / 33
+            - np.cos(35 * ap) / 35
+        )
```
rodrigomoliveira1, line 479: Maybe you could replace all these code lines by doing this:
```python
size = 18
cosine_sum = np.sum([(-1) ** i * np.cos((2 * i + 1) * ap) / (2 * i + 1) for i in range(size)])

kee = kme + (4 / np.pi) * kde * cosine_sum 
knn = kmn + (4 / np.pi) * kdn * cosine_sum 
```
You get a cleaner code and better control over how many terms you would like in your sum, without needing "ctrl+c / ctrl+v each time you add a new term.

ross/defects/crack.py
```diff
@@ -0,0 +1,509 @@
...
+        aux = np.array([[kee, 0], [0, knn]])
+
+        K = ((self.T_matrix.T).dot(aux)).dot(self.T_matrix)
+
+        return K
+
+    def _get_coefs(self, coef):
+
+        cwd = os.getcwd()
+        x = scipy.io.loadmat(cwd + "/tools/data/PAPADOPOULOS_c")
```
rodrigomoliveira1, line 503: In ROSS, the pattern is using `Path` library to manage paths (when saving or loading).
So, you could replace `os` and `sys` by `Path` and these lines by:
```python
cwd = Path(__file__).parent / "tools/data/PAPADOPOULOS_c"
x = scipy.io.loadmat(cwd)
```

Besides, using, `scipy.io.loadmat` intends to load a `MATLAB` file. I don't know if it's the best idea to work with. If it's possible, I'd recommend loading and .CSV file or something more "general".

rodrigomoliveira1 (changes requested):
Hello, @bela-bs. I'm doing some comments to help improving the work.

ross/defects/crack.py
bela-bs: Thank you Rafael, I will make the necessary changes.

ross/defects/crack.py
bela-bs: Yes Rodrigo, it is an excellent alternative indeed!

ross/defects/crack.py
bela-bs: Thanks for the tips! And I will change the file (.mat) format. It is even more convenient.

rodrigomoliveira1 (approved):

ross/rotor_assembly.py
```diff
@@ -2512,6 +2512,43 @@ def run_rubbing(self, **kwargs):
...
+        Examples
+        --------
+        >>> from ross.defects.crack import crack_example
+        >>> probe1 = (14, 0)
+        >>> probe2 = (22, 0)
+        >>> response = crack_example()
+        >>> results = response.run_time_response()
+        >>> fig = response.plot_dfft(probe=[probe1, probe2], range_freq=[0, 100], yaxis_type="log")
+        >>> # fig.show()
+        """
```
raphaeltimbo, line 2551: It looks like `run_crack` is not used in the example.
Also, I recommend adding the `@check_units` decorator to the function, this way the user will be able to pass pint quantities to the function. The way it works is detailed in the `@check_units` docstring:

```
    """Wrapper to check and convert units to base_units.
    If we use the check_units decorator in a function the arguments are checked,
    and if they are in the dictionary, they are converted to the 'default' unit given
    in the dictionary.
    For example:
    >>> units = {
    ... "L": "meter",
    ... }
    >>> @check_units
    ... def foo(L=None):
    ...     print(L)
    ...
    If we call the function with the argument as a float:
    >>> foo(L=0.5)
    0.5
    If we call the function with a pint.Quantity object the value is automatically
    converted to the default:
    >>> foo(L=Q_(0.5, 'inches'))
    0.0127
    """
```
Arguments such as `unbalance_magnitude` are already available in the units dictionary (see [here](https://github.com/ross-rotordynamics/ross/blob/fde96c36fbb087551a2dd78cec25f315f0243c10/ross/units.py#L20-L41)), so I recommend that you just change the argument name from `massunb` to `unbalance_magnitude`.
Other arguments can be added to the units dictionary, but I recommend that you first look to see if you can use one argument name that is already there so we can keep the user interface more simple.

bela-bs: Hi Raphael, do you think that we could open an issue to fix this '@check_units' and change the argument name from massunb to unbalance_magnitude  in all defect classes after we close this PR?

raphaeltimbo: Yes, no problem.

raphaeltimbo:
Investigating the appveyor error, I have run the test_crack.py file in a Windows machine and got this diff from actual to expected values:
image

I believe this could be just something numerical, since the matrix has very high values (e+9) and also very low (e-11).
What is the purpose of this matrix? Maybe we could develop some method to test it based on the most important values from the matrix.

rodrigomoliveira1:
> I believe this could be just something numerical, since the matrix has very high values (e+9) and also very low (e-11).
> What is the purpose of this matrix? Maybe we could develop some method to test it based on the most important values from the matrix.

Or we could test some output the comes after it (I don't exactly know if this matrix is the actual output or just some calculation step).

bela-bs:
We are trying to use another matrix (time-independent) to resolve this issue. However, we are encountering some difficulty to find a parameter that describes correctly the crack model.

@raphaeltimbo @rodrigomoliveira1

bela-bs:
image

We got this error in app veyor now. Do you have any idea what that could be?

@raphaeltimbo @rodrigomoliveira1

rodrigomoliveira1:
@bela-bs, don't worry about this error. It was my fault while trying to make appveyor work with codecov. You can just ignore it and focus on the tests themselves.

raphaeltimbo:
Hi there @bela-bs @verg1lio ! 
I am trying to fix the conflicting files here in this PR, but to do this I need to push to your branch.
Can you check if you guys checked this box?
image

verg1lio:
Hi @raphaeltimbo, doing it now.

bela-bs:
Hi @raphaeltimbo, I don't have this option here:
image

raphaeltimbo:
Hi @bela-bs !
I was looking for a solution to this and I found this discussion: https://github.community/t/how-can-we-enable-allow-edits-from-maintainers-by-default/2847/2

It looks like it is not possible to enable that if the fork is not under your user.

Could you guys solve the conflict in the ` docs/examples/example_13.ipynb ` file?
Basically you will have to pull the latest master, merge and solve the conflicts.

The other option would be to add me to the LMEst-Rotor team so I could do that.
I think it would be better that in the future you can have forks that are user owned to enable this possibility here in the PRs.

raphaeltimbo:
@bela-bs , I have just noticed that you have already sent me an invitation to LMEst.
I have already pushed the merge commit. Let's see how tests go and then we can merge the PR.

bela-bs:
Hi @raphaeltimbo, we have just resolved the latest conflicts, now it's possible to merge this PR. And we will create the PR related to the ``check_units`` decorator as you said earlier.

PR #703, opened 2020-11-23
Add save() and load() methods to manage ross results

rodrigomoliveira1:
Adds an abstract class `ST_Results` that holds a generic `save()` and `load()` methods.
The other classes will inherit these methods from `ST_Results`. It avoids redundancies by repetition of functions with the same functionality.
The user will be able to save the results of a given analysis and load it later, avoiding the need to re-run the entire analysis, which could be costly.

Example of how to use it:

```python
        >>> # Example running a stochastic unbalance response
        >>> from tempfile import tempdir
        >>> from pathlib import Path
        >>> import ross.stochastic as srs

        >>> # Running an example
        >>> rotors = srs.st_rotor_example()
        >>> freq_range = np.linspace(0, 500, 31)
        >>> n = 3
        >>> m = np.random.uniform(0.001, 0.002, 10)
        >>> p = 0.0
        >>> results = rotors.run_unbalance_response(n, m, p, freq_range)

        >>> # create path for a temporary file
        >>> file = Path(tempdir) / 'results.toml'
        >>> results.save(file)

        >>> # Loading file
        >>> results2 = srs.ST_ForcedResponseResults.load(file)
        >>> results2.magnitude.all() == results.magnitude.all()
        True
```

It is first applied to Stochastic ROSS. But the idea will be extended to ROSS.

PR #711, opened 2020-12-01
Add save() and load() methods to manage ROSS results

rodrigomoliveira1:
This PR is a remake from PR #703. One can check the discussion history by accessing the previous link. 

Adds abstract classes "Results" and "ST_Results" that holds a generic save() and load() methods.
The other classes will inherit these methods from "Results" and "ST_Results". It avoids redundancies by repetition of functions with the same functionality.
The user will be able to save the results of a given analysis and load it later, avoiding the need to re-run the entire analysis, which could be costly.

Example of how to use it:

```python
        >>> # Example running a stochastic unbalance response
        >>> from tempfile import tempdir
        >>> from pathlib import Path
        >>> import ross.stochastic as srs

        >>> # Running an example
        >>> rotors = srs.st_rotor_example()
        >>> freq_range = np.linspace(0, 500, 31)
        >>> n = 3
        >>> m = np.random.uniform(0.001, 0.002, 10)
        >>> p = 0.0
        >>> results = rotors.run_unbalance_response(n, m, p, freq_range)

        >>> # create path for a temporary file
        >>> file = Path(tempdir) / 'results.toml'
        >>> results.save(file)

        >>> # Loading file
        >>> results2 = srs.ST_ForcedResponseResults.load(file)
        >>> results2.magnitude.all() == results.magnitude.all()
        True
```

### DIFF
--- a/ross/__init__.py
+++ b/ross/__init__.py
@@ -8,6 +8,7 @@ from .defects import *
 from .disk_element import *
 from .materials import *
 from .point_mass import *
+from .results import *
 from .rotor_assembly import *
 from .shaft_element import *
 from .utils import visualize_matrix
--- a/ross/results.py
+++ b/ross/results.py
@@ -3,8 +3,13 @@
 This module returns graphs for each type of analyses in rotor_assembly.py.
 """
 import copy
+import inspect
+from abc import ABC
+from collections.abc import Iterable
+from pathlib import Path
 
 import numpy as np
+import toml
 from plotly import graph_objects as go
 from plotly.subplots import make_subplots
 from scipy import linalg as la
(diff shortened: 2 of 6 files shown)


raphaeltimbo (approved):

rodrigomoliveira1:
I'll work on a test file to these classes. I need a guarantee that all analyzes are saved and loaded correctly. Please, don't merge it yet.

rodrigomoliveira1:
@raphaeltimbo, 
I had to remove the argument `lti` from `ModalResults`  because it was not trivial to save a `StateSpaceContinuous` and it was not being used anywhere else in ModalResults.

It was called in `time_response` like `modal.lti`, but it was unnecessary to run the modal analysis to get the `lti`. So, I changed it to call directly the method itself `lti = self._lti(speed)` instead.

Issue #714, opened 2020-12-02
Add major and minor axis in response plots


raphaeltimbo:
As mentioned by @leoishimoto in #710:

> The major and minor axis of the ellipse is highly desirable not only the probes reference. In circular orbits does not make difference but for anisotropic bearings is necessary.

@rodrigomoliveira1, you have mentioned that you are already working on a solution for this one, so I am going to assign you here.

rodrigomoliveira1:
As requested by @LeifParr, 

> Another comment regarding the response plotting, is that the plots should indicate the node number (and the label) which are plotted. As far as I can see, there is no option to label probes, so there is room for confusion about what is being shown. If you consider this an important issue, you can move this into a new issue.

I'm including optional tags to name `probes` and the node number. If the probe tuple has only 2 items (node and orientation), the tag is set to a default name, like it was previously ("Probe 1", "Probe 2", etc)
Example:

```python
>>> rotor = rs.rotor_example()
>>> speed = np.linspace(0, 1000, 301)
>>> response = rotor.run_unbalance_response(
...                                         node=3,
...                                         unbalance_magnitude=0.001,
...                                         unbalance_phase=0.0,
...                                         frequency=speed)

>>> # POST-PROCESSING
>>> # Get results from node 3 with 90° orientation
>>> probe_node_1 = 3
>>> probe_angle_1 = np.pi / 4

>>> # Get results from node 3 for major axis
>>> probe_node_2 = 3
>>> probe_angle_2 = "major" # here we insert the string "major" to get the major axis values

>>> # Get results from node 3 for minor axis
>>> probe_node_3 = 3
>>> probe_angle_3 = "minor" # here we insert the string "minor" to get the minor axis values
>>> response.plot_magnitude(probe=[
...     (probe_node_1, probe_angle_1),
...     (probe_node_2, probe_angle_2, "My Tag 1"),
...     (probe_node_3, probe_angle_3, "My Tag 2"),
... ], yaxis=dict(type="log"))
```
![image](https://user-images.githubusercontent.com/45969994/101013374-f2a5d500-3542-11eb-8b30-3b44f01a1482.png)

PR #716, opened 2020-12-03
Add minor and major axes to response plots

rodrigomoliveira1:
- New method in `ForcedResponseResults`: `_calculate_major_axis_per_node()` calculates the major and minor axes at a given node for all frequencies in `speed_range` array. Plotting methods will use this function to simplify and clear their codes.
- New option to add tags to probes in unbalance and time response. Users Will be able to name probes. If nothing is passed, it uses a default tag enumerating the probes.

An example of how these features works is shown in Issue #714.

### DIFF
--- a/ross/results.py
+++ b/ross/results.py
@@ -1467,7 +1467,7 @@ class ForcedResponseResults(Results):
             angle = Q_(p[1], probe_units).to("rad").m
             vector = self._calculate_major_axis_per_node(node=p[0], angle=angle)
             try:
-                probe_tag = p[2] + f" - Node {p[0]}"
+                probe_tag = p[2]
             except IndexError:
                 probe_tag = f"Probe {i+1} - Node {p[0]}"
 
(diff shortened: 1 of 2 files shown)


ross/results.py
```diff
@@ -1458,38 +1459,26 @@ def plot_magnitude(
...
-
-            probe_resp = operator @ np.vstack((magnitude[dofx], magnitude[dofy]))
-            z = np.sqrt((probe_resp[0] * np.cos(angle)) ** 2 +
-                        (probe_resp[1] * np.sin(angle)) ** 2)
-            # fmt: on
-
-            z = Q_(z, "m").to(amplitude_units).m
+            vector = self._calculate_major_axis_per_node(node=p[0], angle=angle)
+            try:
+                probe_tag = p[2] + f" - Node {p[0]}"
```
raphaeltimbo, line 1470: If the user wants to give a specific tag to the probe I think it is not a good idea to modify it, so I would remove the `f" - Node {p[0]}"` part, since this could be added by the user.

raphaeltimbo (approved):

raphaeltimbo:
@rodrigomoliveira1 , this is really good!
Just made a small comment regarding the probe tag. I was thinking about cases where the user may want only a word like 'Center' or 'Bearing' and the node information could become a pollution in the plot. But I don't mind to merge it how it is now and latter we could check others opinions.

ross/results.py
rodrigomoliveira1: Done in commit [4b3586b](https://github.com/ross-rotordynamics/ross/pull/716/commits/4b3586b3fcd42c2d2732fb1d9d0f46e035a29f8c).

PR #731, opened 2020-12-16
Updating docs and website

rodrigomoliveira1:
- Update stochastic ROSS tutorial
- Fix links for downloading examples and tutorials
- Add reference to our paper published in JOSS at the index page.


rodrigomoliveira1:
@raphaeltimbo 
let's use this PR to update our docs at the website. I see there's still things to do like:

- Fix example 11 title - There're 2 Examples 10
- Add example 12 to the examples page
- Fix examples 10 and 11 outputs, which are not being displayed at the page.
- Remove "Tutorial" from the tab on the top of the page - We had only 1 tutorial file, now we got 3.
- Anything else you remember

verg1lio:
@rodrigomoliveira1 that time was needed for the rotor system to reach its dynamic steady-state, specially in the examples. I believe @bela-bs can explain it better. We did try using smaller time intervalos, but as i remember it didn't stabilized enough for the test comparisons to be valid.

rodrigomoliveira1:
Here are the outputs in Examples 10, 11 and 12:

Example 10:
newplot

Example 11
newplot(1)

Example 12
newplot(2)

We can still run the tests setting a larger time range, I changed just the examples to decrease their sizes.

verg1lio:
> Here are the outputs in Examples 10, 11 and 12:
> 
> Example 10:
> newplot
> 
> Example 11
> newplot(1)
> 
> Example 12
> newplot(2)
> 
> We can still run the tests setting a larger time range, I changed just the examples to decrease their sizes.
> 
> 

Ah, perfectly fine then. It took us a while to realize that the simulation time was the problem for the tests convergence, but for the tutorial it should be fine. Thanks for the fix!

raphaeltimbo (approved):

PR #732, opened 2020-12-17
Add function to collect and organize data from figures

rodrigomoliveira1:
This function takes a `go.Figure()` object from Plotly and converts the data to a `DataFrame` from Pandas. It works only for Scatter Plots (XY) and does no support subplots or polar plots.

This function is suitable to the following analyzes:
- `run_freq_response()`
- `run_unbalance_response()`
- `run_time_response()`


raphaeltimbo (approved):
I think this is a good thing to have, but I am wondering if this data should already be available as an attribute inside the results object.

rodrigomoliveira1:
> but I am wondering if this data should already be available as an attribute inside the results object.
 I think it's a better idea.

PR #743, opened 2021-01-22
Add Black, Isort and Codecov to workflow

rodrigomoliveira1:
Functionalities to add to workflow:
- [x] Black
- [x] Isort
- [x] Codecov
- [x] Deploy docs 

Adding some routines to the workflow to run after the tests.


raphaeltimbo:
It looks like secrets are only available if we create a PR from a branch within the repository (not from a fork).
I will merge this one and try to create a branch inside the repository to test it.

PR #744, opened 2021-01-25
THD Cylindrical Bearing

marcusfilipesr:
This PR presents the thermohydrodinamic code for a cylindrical bearing. It is passing through validations with the MaxBrg results, although, plenty of improvements has to be made, so the class is fitted with the actual ross structure. 

@verg1lio @raphaeltimbo @rodrigomoliveira1 @elyqg @bela-bs


raphaeltimbo (changes requested):

elyqg:
Hi @raphaeltimbo. We are correcting some unexpected issues in the code and this week we will start making all these corrections that you pointed.

marcusfilipesr:
Hi @raphaeltimbo, we were checking the differences between our class' attributes and the `fluid_flow`'s, and we have some doubts around a few variables that you have implemented, on how to use/generate them. We would like to know if it is possible for us to appoint a meeting with you guys, so then we can discuss an optimum solution.

We'll also copy this comment on the issue so everyone can see it.

elyqg:
Hi @raphaeltimbo. We made the corrections according to your recommendations and also added the documentation to the code. We would like you to review this part and give us a feedback. We are still in the final stage of validation but we take the opportunity to advance this part of the documentation.

raphaeltimbo:
Hi there @elyqg !
Just some general comments, the file is still importing matplotlib which is not part of ROSS requirements, so this causes the tests to fail at import time.
For now, I think the most important thing to do is to add tests and remove the matplotlib dependence.
I recommend moving the script part of the code (everything after `if __name__ == '__main__'`) to a test file.
You can create a new test file like `test_thdcylindrical.py` or use the `test_fluid_flow.py` file as you like.

elyqg:
Hi @raphaeltimbo . We made the corrections mentioned and implemented the testing part. The tests are slow and heavy due mainly to the mesh size. We are correcting some errors in the code to use a smaller one. We had problems with the codecov as well. If you can help us later because we don't know how to fix it.

ross/tests/test_cylindrical.py
```diff
@@ -0,0 +1,153 @@
...
+        Cp,
+        rho,
+        Treserv,
+        mix,
+        T_1,
+        T_2,
+        mu_ref1,
+        mu_ref2,
+    )
+    mancal.run(x0, print_progress=True)
```
raphaeltimbo, line 60: @elyqg , I think you could remove this part, since it is already present at each pytest fixture.

marcusfilipesr: We'll do it right away!

elyqg:
Hi @raphaeltimbo . @marcusfilipesr corrected the errors that you pointed out. If you see some other points just inform that we will correct them. We are still validating the results and there may be some changes to the code, but nothing that changes its structure.

raphaeltimbo:
Hi @elyqg ! 
I saw that you have removed the `bearing.run()` from the `cylindrical_units` fixture, but I would also remove that from the cylindrical fixture and the call it inside the tests that really need that:

```python
def test_cylindrical_equilibrium_pos(cylindrical):
    cylindrical.run([0.1, -0.1])
    assert math.isclose(cylindrical.equilibrium_pos[0], 0.58656872, rel_tol=0.01)
    assert math.isclose(cylindrical.equilibrium_pos[1], -0.67207557, rel_tol=0.01)
```

The `test_cylindrical_parameters` would remain the same, but it will run faster since we don't need to run for that test.

elyqg:
Hi @raphaeltimbo. The THD cylindrical bearing code is in the final stage of validation. The tests are running and I believe that only small details will be corrected from now on. Is there something else to make on it or we can proceed with the merging?
Thanks, and have a great day!

raphaeltimbo (approved):

raphaeltimbo:
Hi @elyqg !
I am going to merge this code now.
The next step would be to implement a THDCylindrical that inherits from BearingElement. 
This way we will be able to add this kind of bearings to our rotor models.

PR #745, opened 2021-01-27
Try to deploy from a branch within ross repository

raphaeltimbo:
Try to deploy from a branch within ross repository.


raphaeltimbo:
So this worked.
When using github secrets we cannot run the action in a PR from a fork.

PR #751, opened 2021-02-02
Remove old CI files

raphaeltimbo:
This removes some CI files that are not used anymore after our
migration to github actions. For reference:
- .travis.yml: travis configuration file
- appveyor.yml: appveyor configuration file
- deploy_docs.sh: script to deploy the docs using travis and the
ross-bott account.
- deploy_key.enc: This was used by travis CI as an encryption to the
ross-bott account ssh key.


rodrigomoliveira1:
We could also remove the badges from README.md and replace them by GitHub actions

raphaeltimbo:
The appveyor check for this PR is failing because I have removed the config file.
I have just now removed appveyor also from the settings in our repository, so I believe we won't have this problem in future PRs.

rodrigomoliveira1 (approved):

PR #752, opened 2021-02-08
Include FluidFlow reference.

JuliaMota:


ross/fluid_flow/fluid_flow.py
```diff
@@ -177,6 +177,10 @@ class FluidFlow:
     numerical_pressure_matrix_available: bool
         True if numerically calculated pressure matrix is available.
 
+    References
+    ----------
+    The complete FluidFlow theory can be found at :cite:`mota2020`
+
```
rodrigomoliveira1, line 183: This section should be about the reference itself. So, you can add just the code to generate the references.

```python
    References
    ----------
    .. bibliography:: ../../../docs/refs.bib
```

This sentence ("The complete FluidFlow theory can be found at :cite:`mota2020`") could be move to the top of the docsting, like:

```python
class FluidFlow:
    r"""Generate dynamic coefficients for hydrodynamic bearings.

    This class calculate the pressure matrix and the stiffness and damping matrices
    of a fluid flow with the given parameters.
    It is supposed to be an attribute of a bearing element,
    but can work on its on to provide graphics for the user.

    The complete FluidFlow theory can be found at :cite:`mota2020`.
    """
```

rodrigomoliveira1 (changes requested):

rodrigomoliveira1 (approved):

ross/fluid_flow/fluid_flow.py
JuliaMota: I agree. I'll do it.

PR #756, opened 2021-02-15
Check_units on defects' class

marcusfilipesr:
As @raphaeltimbo requested in another PR, we (@bela-bs and I) created this one to apply the `check_units` decorator. If you guys have any comments of what more we can improve, we'll be delighted to hear.


raphaeltimbo:
Hi there @marcusfilipesr and @bela-bs !
I have pushed some commits to this PR to give some examples on the crack file.
If you could do the same modifications to the misalignment file (improve docstrings, add units to other arguments when possible, and add tests).

bela-bs:
Hi @raphaeltimbo, we just finished the changes according to the units and names of the variables as requested. 
Do you have any other improvements that you would like us to make?

ross/defects/crack.py
```diff
@@ -30,8 +30,8 @@ class Crack(Defect):
         Initial time
     tF : float
         Final time
-    depth : float, pint.Quantity
-        Crack depth. Default unit is meter.
+    depth_ratio : float
+        Crack depth ratio. A depth value of 0.1 is equal to 10%, 0.2 equal to 20%, and so on.
```
raphaeltimbo, line 34: Is this ratio related to the shaft diameter?

raphaeltimbo:
Hi @bela-bs !
My only question is regarding the depth_ratio. 
I would assume that the ratio is related to the shaft diameter, since within ROSS we are always defining shaft elements in terms of diameter. But I saw that you have defined an attribute `radius` within the crack class, so I am not sure if the depth_ratio is related to radius or diameters. It would be nice if you guys could make that explicit within the docstrings.

bela-bs:
Of course @raphaeltimbo , this depth_ratio variable is related to the diameter of the element that contains the crack, we will make this clear in the documentation.

raphaeltimbo (approved):

PR #762, opened 2021-02-19
Change methods to `run_`

raphaeltimbo:
This PR changes the `plot_level1` and `plot_ucs` methods so that they behave as other run methods.
I have added the `run_` prefix and created results classes to store data / methods as we do in other cases.


raphaeltimbo:
I will close this one to fix the commit history.

PR #765, opened 2021-02-19
Fix errors in example 7

rodrigomoliveira1:
- Updates command line calling `BearingFluidFlow`.
- Fix bug in `Rotor` class where `y_pos_sup` would not be calculated for certain bearing types.


ross/rotor_assembly.py
```diff
@@ -483,11 +487,14 @@ def flatten(l):
...
         # define position for point mass elements
+        # fmt: off
         dfb = df[
-            (df.type == "BearingElement")
-            | (df.type == "BearingElement6DoF")
-            | (df.type == "SealElement")
+            df.type.isin([
+                "BallBearingElement", "BearingElement", "BearingFluidFlow",
+                "BearingElement6DoF", "MagneticBearingElement", "RollerBearingElement",
+                "SealElement"])
```
raphaeltimbo, line 495: To avoid having to edit this in the future, maybe we could do something like:
```python
dfb = df[df.apply(lambda x: isinstance(x, BearingElement))]
```

raphaeltimbo (commented):
See comment regarding type check.

ross/rotor_assembly.py
rodrigomoliveira1: It doesn't work.
`df.type` is column of `str` types, so it always returns `False`

rodrigomoliveira1: I found another way, but we need to import `sys` and `inspect`.

```python
classes = [_class for _class, _ in inspect.getmembers(sys.modules["ross.bearing_seal_element"], inspect.isclass)]
dfb = df[df.type.isin(classes)]
```

raphaeltimbo: This is good for me!

rodrigomoliveira1: i can replace some unused imports on `rotor_assembly.py` like `os` and `shutil`

raphaeltimbo (approved):

PR #771, opened 2021-02-24
WIP - Converting CoAxialRotor to MultiRotor class

rodrigomoliveira1:
- Add "n" as property to Element Class
  - This feature allows us to manipulate the node value of any element even after building a rotor. It'll make the task of changing nodes easier in MultiRotor class.
  - Changes all references of "n" to "_n".

- More coming soon


ross-bott:
Hi there!
less than 45 days;

PR #777, opened 2021-02-26
Fix defects examples

bela-bs:
Hi @raphaeltimbo, we found some errors from the last PR, related to the variable `speed` and other minor errors present in defects' examples. This PR is being created to solve these problems.

### DIFF
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -1,5 +1,9 @@
 name: Release
-on: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
+on:
+  push:
+    tags:
+      - '*'
+
 jobs:
   build-n-publish:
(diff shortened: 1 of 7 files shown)


raphaeltimbo (changes requested):
Hi there @bela-bs !
Regarding the changes to `abs_defect.py` and `integrate_solve.py` I don't have any comments.
For the changes in the examples, I could not see everything that has changed since the git diff for the notebooks is not so good, but I saw that you have added a cell at the start with the following:

```python
import sys

from pathlib import Path

sys.path.append(str(Path(os.getcwd()).parent.parent))
```

This actually is not necessary for the user that has ROSS installed.
Could you please revert these changes before we merge it?

Thanks!

bela-bs:
@raphaeltimbo Ah yes, I just did it.

raphaeltimbo (approved):

PR #806, opened 2021-06-16
Add sphinx copy button to conf.py

raphaeltimbo:
The sphinx copy button is now added to the setup.py and to the sphinx
conf.py, so it is not necessary to have its source code in our files.


raphaeltimbo:
Close for now. Will migrate to readthedocs first.

Issue #814, opened 2021-06-30
Fix - 4 and 6 DoF EIG solution method modification


verg1lio:
Issue related to the numerical precision of the solver method adopted for the eigenvalue problem, aiming to change the solver for a more precise and robust one. The current solver works fine for small matrices (smaller than ~30x30), but struggles to keep the error acceptable for anything bigger than that. Some models demand bigger and more poorly conditioned matrices, which demand a different solver.

raphaeltimbo:
Hi there @verg1lio !
The function used to calculate eigenvalues basically has this in it: https://github.com/ross-rotordynamics/ross/blob/f7cc3d118a727313bf648805b1b5e72b610e3f6d/ross/rotor_assembly.py#L1237-L1251

So we are using the scipy [eigs](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.eigs.html) function, which uses the Fortran [ARPACK](https://www.caam.rice.edu/software/ARPACK/) to calculate the eigenvalues. 

> ARPACK is a collection of Fortran77 subroutines designed to solve large scale eigenvalue problems.

As you can see in the code we have to set some arguments for the calculation of the eigenvalues when calling the `la.eigs()` function. I think we should start this investigation by checking if any of these arguments are having a negative impact on the calculation precision. 

We also need to discuss what level of precision we want here, since changing the solver may affect calculation time and also can increase our dependencies, which could increase package installation size, time to build and test the software etc. So we will need to balance those aspects as well.

To evaluate this, I would recommend that we select some study cases (models) to use as reference when comparing the solvers in precision and calculation time. If you could share the models that you are seeing a larger error here we could start by those.

verg1lio:
Hello @raphaeltimbo !

I start by agreeing with you here  _"As you can see in the code we have to set some arguments for the calculation of the eigenvalues when calling the la.eigs() function."_ , since if this routine already uses sufficiently accurate enough eigenvalue solution methods it is a simple method of using the right parameters.

[...]

PR #826, opened 2021-07-15
Fix eig method

verg1lio:
Pull request related to [https://github.com/ross-rotordynamics/ross/issues/814](url).

This is a simple fix to a persistent error related to test functions using the default eigensolver from `scipy.linalg.eig`, which uses different methods for sparse and full matrices, and also presents different methodologies based on matrix size and singularity factor. It is now set to run as `sparse=False` in all test cases. It should provide the necessary precision to the solver for its answers to be reliably compared to test cases, and then for them to pass without further issues.


verg1lio:
image

I am still having some precision errors for some reason... will try updating the benchmark solution results with the new `sparse=False` option to see if it matches the current solution. Some other tests are failing too, despite not having been modified here.

verg1lio:
Problem solved and test function for 6dof model is passing. Other unrelated test functions are falining, and should be fixed in appropriated PRs.

image

verg1lio:
> Problem solved and test function for 6dof model is passing. Other unrelated test functions are falining, and should be fixed in appropriated PRs.
> 
> image

Either that, or I do not have some of the requirement packages installed locally when running the tests.

ross/rotor_assembly.py
```diff
@@ -3821,10 +3821,21 @@ def rotor_example_6dof():
     >>> import ross as rs
     >>> import numpy as np
     >>> rotor6 = rs.rotor_assembly.rotor_example_6dof()
-    >>> camp6 = rotor6.run_campbell(np.linspace(0,400,101),frequencies=18)
 
-    # plotting Campbell Diagram
+    Plotting rotor model
```
raphaeltimbo, line 3825: Hi @verg1lio ! Going [here](https://github.com/ross-rotordynamics/ross/pull/826/checks?check_run_id=3086398825) you can see what tests are failing in the PR.
Notice that we also test the examples in the docstrings. We do this to be sure that all code provided in the examples are always working.
The rules for this part of the docstrings are: 
- Everything with '>>>' is code that will be executed;
- Everything that does **not** have '>>>' is output, and this value will be compared with the code output for the test to pass;
- All comments should have '#' to avoid execution.
Here the text 'Plotting rotor model' does not have the '#', since this is not valid code the test will fail.

ross/rotor_assembly.py
```diff
@@ -3821,10 +3821,21 @@ def rotor_example_6dof():
...
+    >>> fig = rotor6.plot_rotor()
+    >>> fig.show()
```
raphaeltimbo, line 3827: We should also remove `fig.show()` from the docstring examples, since the output representation of the figure is something that is hard to keep track of. I recommend just inserting a comment with '# to plot use fig.show()'.

ross/tests/test_rotor_assembly.py
```diff
@@ -1557,27 +1557,34 @@ def rotor_6dof():
     return Rotor(shaft_elem, [disk0, disk1], [bearing0, bearing1])
 
 
-@pytest.mark.skip(
-    reason="Needs investigation. It fails depending on system. Most likely due to eig solution precision"
-)
+# @pytest.mark.skip(
+#   reason="Needs investigation. It fails depending on system. Most likely due to eig solution precision"
+# )
```
raphaeltimbo, line 1562: Since we are going to use the test, I think we could just remove the skip instead of just commenting it.

ross/rotor_assembly.py
```diff
@@ -3821,10 +3821,21 @@ def rotor_example_6dof():
...
+
+    Running modal
+    >>> rotor_speed = 100.0 # rad/s
+    >>> modal6 = rotor6.run_modal(rotor_speed)
+    >>> print(f"Undamped natural frequencies:\n {modal6.wn}")
+    >>> print(f"Damped natural frequencies:\n {modal6.wd}")
```
raphaeltimbo, line 3833: You also have to provide the output here for these prints, otherwise the test will fail.
Also, to avoid having to write big arrays in the docstrings you can use ellipsis to shorten these arrays.
Here is an example:
```python
    >>> bearing0.K(frequency) # doctest: +ELLIPSIS
    array([[[1000000., 1000000., ...
    >>> bearing0.C(frequency) # doctest: +ELLIPSIS
    array([[[200., 200., ...
```

raphaeltimbo (changes requested):
Hi @verg1lio ! 
I did some comments regarding the docstrings that were altered in this PR and are causing the tests to fail.
If you want to test these docstrings locally, you can run `pytest ross/rotor_assembly.py` in the terminal to check it.
If you have questions let me know!

raphaeltimbo:
Hi @verg1lio ! 
Any updates for this PR?
Let me know if you need any help to fix the doctests.

verg1lio:
> Hi @verg1lio ! 
> Any updates for this PR?
> Let me know if you need any help to fix the doctests.

Hello @raphaeltimbo , good evening!

I believe that the tests that are failing to pass are unrelated to the main topic of this PR. That being the case, shouldn't i open specific PRs for them?

If you wish I can solve them here, no problems.

Cordially, 
Vergilio

raphaeltimbo:
@verg1lio, I believe some tests are failing here because of a modification in the rotor_example_6dof docstring.
The ross repository is configured to block merging code that is not passing the tests, so if you could please fix these, it would be great.
Thank you!

verg1lio:
Ah alright then. I will fix the failing issues in this PR and notify when its done.

Thanks for the heads up!

verg1lio:
Dear @raphaeltimbo , good afternoon!

I have been trying to pinpoint the errors, but without success. Running the tests locally they pass without any issue. See images below.

image
image

That being said, the tests run online evidently do not pass... maybe I made some modification to the rotor_example_6dof docstring, but I do not seem to find any problem with it. I will consult with my coworkers for the moment and get back to you when the issue is solved.

Have a great afternoon!
Vergílio

raphaeltimbo:
Hi @verg1lio ! I will try to reproduce this in my machine and see if I can commit some code to fix the error.
I will let you know if I have some progress in this.

raphaeltimbo:
Hi @verg1lio! I have made some modifications to the docstring in commit https://github.com/ross-rotordynamics/ross/pull/826/commits/2aafb46b9ffd35dbecb218b8b70aa6dc8355f521.
I will merge this PR now.
Thank you!

verg1lio:
Thanks greatly for the assistance, and have a good evening!

PR #872, opened 2022-02-06
Fix lacking output in Level1Results

rodrigomoliveira1:
.plot() method did not have a return command.


raphaeltimbo (approved):
Hi @rodrigomoliveira1. Thank you for the contribution.
CI is failing, but I think it is because of a new black version (22). Will fix this in a different PR.

Will merge this one now.

Issue #874, opened 2022-02-19
Bug while plotting bearing coefficients


rodrigomoliveira1:
Hello guys,

Recently, I was running some analyzes and I found an issue while plotting multiple bearing coefficients.

I will give you a rapid example of what's happening:

```python
import ross as rs

bearing_file = (
     os.path.dirname(os.path.realpath(__file__)) + "/data/bearing_seal_si.xls"
)

bearing = BearingElement.from_table(0, bearing_file)

fig = bearing .plot(["kxx", "kyy"], "rpm")
fig.show()
```
![image](https://user-images.githubusercontent.com/45969994/154782605-e5f49655-17b2-45d9-ac1b-e2db2d7cb1ba.png)

Doing some tests, I've noticed it only happens if you are trying to converting units. It's ok if u plot it in rad/s, look:

![image](https://user-images.githubusercontent.com/45969994/154782654-65ff8d67-b117-48a7-a768-d10a9ca11a8b.png)

So, the problem is the orinal `frequency_range` variable is overwritten every time the code needs to convert it to plot a new coefficient. It converts `rad/s` to `rpm` in the fisrt run, then it does the same thing again for the following coefficients. A small change, letting the original variable as an auxiliar should fix it:

From:
```python
        frequency_range = np.linspace(min(self.frequency), max(self.frequency), 30)

        for coeff in coefficients:
            y_value = (
                Q_(
                    getattr(self, f"{coeff}_interpolated")(frequency_range),
                    default_units,
                )
                .to(y_units)
                .m
            )
            frequency_range = Q_(frequency_range, "rad/s").to(frequency_units).m

            fig.add_trace(
                go.Scatter(
                    x=frequency_range,
                    y=y_value,
                    mode="lines",
                    showlegend=True,
                    hovertemplate=f"Frequency ({frequency_units}): %{{x:.2f}}<br> Coefficient ({y_units}): %{{y:.3e}}",
                    name=f"{coeff}"
                )
            )
```

To:
```python
        _frequency_range = np.linspace(min(self.frequency), max(self.frequency), 30)

        for coeff in coefficients:
            y_value = (
                Q_(
                    getattr(self, f"{coeff}_interpolated")(_frequency_range),
                    default_units,
                )
                .to(y_units)
                .m
            )
            frequency_range = Q_(_frequency_range, "rad/s").to(frequency_units).m

            fig.add_trace(
                go.Scatter(
                    x=frequency_range,
                    y=y_value,
[...]

PR #875, opened 2022-02-19
Fix converting units when plotting multiple coefficients

rodrigomoliveira1:
Related to Issue #874.


raphaeltimbo (approved):
Thanks again for the contribution @rodrigomoliveira1 !

PR #879, opened 2022-03-22
Cylindrical bearing ross standardization

elyqg:
Cylindrical bearing ross standardization


ross/fluid_flow/cylindrical.py
```diff
@@ -205,11 +180,70 @@ def __init__(
...
+                kyy = coef[3]
+
+                stiff = False
+            else:
+                cxx = coef[0]
+                cxy = coef[1]
+                cyx = coef[2]
+                cyy = coef[3]
+
+        super().__init__(node,kxx,cxx,kyy,kxy,kyx,cyy,cxy,cyx)
```
raphaeltimbo: The `super().__init__` call does not have the `frequency` argument here. This way, we can only have a bearing with constant coefficients instantiated. We should have a way to call `THDCylindrical(..., frequency=array`), to calculate the coefficients to different frequencies and pass that to the `super().__init__()` call.

raphaeltimbo (changes requested):

elyqg:
Hi @raphaeltimbo. The proposed corrections have been finalized. If something is still wrong, just let us know and we'll fix it.

raphaeltimbo (approved):

raphaeltimbo:
Merged the code.
Thank you @elyqg and @marcusfilipesr !

PR #907, opened 2022-10-11
Remove setup.py <3.10 version restriction

eu-cristofer:
This change aims to avoid error when installing it in python env version 3.10.6


raphaeltimbo:
I have also added python 3.10 to the test matrix.
Merging this now.
Thank you @eu-cristofer !

PR #910, opened 2022-10-14
Tests corrections

### LINKED ISSUE #818: Cylindrical tests taking too long to complete
The tests in the `test_cylindrical.py` file take a long time to complete. 
I will try to decrease this time by changing the number of elements in the calculation and also by grouping some tests.

### LINKED ISSUE #908: Errors in the misalignment tests
For some reason two misalignment tests are now failing in the github actions:

```
ross/tests/test_misalignment.py ..........FF....
```

These are the `test_mis_angular_forces` and `test_mis_angular_forces_units`.
The tests are failing only with ubuntu and python 3.9:

image

We should check if this is just a problem of precision or if it is necessary to fix something in the code.

### LINKED ISSUE #909: Material test error with python 3.10
With #907, python 3.10 is now supported.
After merging the PR I have noticed that one test is failing for this python version:

```python
________________________________ test_error_rho ________________________________

    def test_error_rho():
        with pytest.raises(TypeError) as ex:
            Material(name="test", E=203.2e9, G_s=80e9)
    
>       assert "__init__() missing 1 required positional argument: 'rho'" == str(ex.value)
E       assert "__init__() m...gument: 'rho'" == "Material.__i...gument: 'rho'"
E         - Material.__init__() missing 1 required positional argument: 'rho'
E         ? ---------
E         + __init__() missing 1 required positional argument: 'rho'
```

marcusfilipesr:
Closes #818
Closes #908 
Closes #909


marcusfilipesr:
Hi @raphaeltimbo! I've created this PR just for the cylindrical, but since I was working with the tests, I've already tried to solve the misalingment and rubbing issue.

It seems like I've solved it, now it is raising an problem with the new version of python 3.10 in the `test_materials.py` file, where the raise is missing the name of the class:

```python
Material.__init__() missing ...
```

ross/tests/test_materials.py
```diff
@@ -55,8 +57,10 @@ def test_specific_material(AISI4140):
 def test_error_rho():
     with pytest.raises(TypeError) as ex:
         Material(name="test", E=203.2e9, G_s=80e9)
-
-    assert "__init__() missing 1 required positional argument: 'rho'" == str(ex.value)
+    if sys.version_info.minor == 10:
+        assert "Material.__init__() missing 1 required positional argument: 'rho'" == str(ex.value)
+    else:
+        assert "__init__() missing 1 required positional argument: 'rho'" == str(ex.value)
```
raphaeltimbo: This solves the problem for version 10, but in the future we would need to add something for version 11...
Since the purpose of this test is just to check if the correct error message is being raised, I would suggest to just change from:

```python
assert "__init__() missing 1 required positional argument: 'rho'" == str(ex.value)
```

To:
```python
assert "__init__() missing 1 required positional argument: 'rho'" in str(ex.value)
```

raphaeltimbo (changes requested):
Hi @marcusfilipesr!
This is looking great. 
Just made one comment regarding the Material test.

ross/tests/test_materials.py
marcusfilipesr: Great! I did not know that this was possible! I'll make this correction right now!

raphaeltimbo:
Thank you @marcusfilipesr!
I've just merged the PR.

Issue #674, opened 2020-09-22
Tutorial links not working


raphaeltimbo:
I tried to use the links for the tutorial inside the ross web page (https://ross-rotordynamics.github.io/ross-website/examples/tutorial.html) and, although they work when we are using the jupyter notebook, it looks like they do not work there on the website.

rodrigomoliveira1:
The funny part is it's a HTML command. It was supposed to work on a website too.

ross-bott:
Hi there!
I have marked this issue as stale because it has not had activity for 45 days.
Consider the following options:
- If the issue refers to a large task, break it in smaller issues that can be solved in
less than 45 days;
- Label the issue as `wontfix` or `wontfix for now` and close it.

ross-bott:
Hi there!
I have marked this issue as stale because it has not had activity for 45 days.
Consider the following options:
- If the issue refers to a large task, break it in smaller issues that can be solved in
less than 45 days;
- Label the issue as `wontfix` or `wontfix for now` and close it.

gutilacerda:
In the ross current website the pages of tutorial and examples are working. Is there another problem related with the website?

raphaeltimbo:
Hi @gutilacerda.
The problem here in this issue is that, in the image bellow, the links in the blue part do work but the links at the red part do not work. 
Here we have some options:
 - Remove the 'Table of Contents' in the red part and stay only with the 'Contents' in the blue part. This has the down side of not having the links when the users are running the notebook on their machines;
 - Figure out a way to make these links work;
 - Try to remove this 'Table of Contents' from the online documentation only, maybe by adding some tag to the notebook cell metadata (we would need to check if [myst-nb](https://myst-nb.readthedocs.io/en/latest/) has this feature).

raphaeltimbo:
Sorry, @gutilacerda. There was some error when I pasted in the message.
I will edit the message and add it.

ross-bott:
Hi there!
I have marked this issue as stale because it has not had activity for 45 days.
Consider the following options:
- If the issue refers to a large task, break it in smaller issues that can be solved in
less than 45 days;
- Label the issue as `wontfix` or `wontfix for now` and close it.

Issue #914, opened 2022-10-28
Argument L is missing in the ShaftElement docstring


raphaeltimbo:
Argument L is missing in the ShaftElement docstring.

raphaeltimbo:
Hi @marcusfilipesr! 
Since this is a very small fix I don't see a problem in doing that.
Just be sure to link that PR with this issue here (write `Closes #914` in the PR message).

ross-bott:
Hi there!
I have marked this issue as stale because it has not had activity for 45 days.
Consider the following options:
- If the issue refers to a large task, break it in smaller issues that can be solved in
less than 45 days;
- Label the issue as `wontfix` or `wontfix for now` and close it.

PR #916, opened 2022-11-04
Fix Table of Contents

gutilacerda:
Hello,

I've changed how the Table of Contents is done, with @marcusfilipesr help. Unfortunately, we could not use Sphinx to build the page from the website to test the modification results. Locally, the HTML given by Sphinx does not build the links in the Table of Contents.


gutilacerda:
The modifications in the first commit called "proportional damping" were removed for this pull request!

marcusfilipesr:
Hi @gutilacerda! When you can, check the issue #914, because we can use this PR to resolve this additional update.

gutilacerda:
Hi @marcusfilipesr. 

Ok, I will check the issue #914 and add the solution in this PR.

gutilacerda:
Hello @marcusfilipesr. 

The correction from issue #914 is already included.

ross/shaft_element.py
```diff
@@ -47,7 +47,8 @@ class ShaftElement(Element):
     odr : float, pint.Quantity, optional
         Outer diameter of the element at the right position (m).
         Default is equal to odl value (cylindrical element).
-    L : shaft length (m)
+    L : float
+        Shaft length (m).
```
marcusfilipesr, line 51: Good morning @gutilacerda! In fact, the `L` parameter can be passed as a `pint.Quantity` as well, similar to `odr`. 

Additionally, as it is the first parameter in the `__init__` method, maybe it is better if it is the first in the documentation.

marcusfilipesr:
Closes #674

ross/shaft_element.py
gutilacerda: Done!

ross/shaft_element.py
```diff
@@ -36,7 +36,8 @@ class ShaftElement(Element):
 
     Parameters
     ----------
-
+    L : float, pint.Quantity
+        Shaft length (m).
```
marcusfilipesr, line 40: One little modification left, instead of `Shaft length` it is the `Element length`.

gutilacerda: Also done, @marcusfilipesr

ross-bott:
Hi there!
less than 45 days;

gutilacerda:
The tables of contents are removed now, only remaining the table on the right side.

raphaeltimbo (approved):

PR #925, opened 2022-11-17
Probe Class

### LINKED ISSUE #827: Add Probe class
Currently we use list of tuples to add probes to our results.
The problem is that, contrary to other parts of ROSS, we are not able to use pint.Quantity to the probe orientation.
Instead, we have to declare the unit in a separate argument:

```
        probe : list
            List with tuples (node, orientation angle, tag).

            node : int -> Indicate the node where the probe is located.

            orientation : float -> Probe orientation angle about the shaft.
            The 0 refers to +X direction.
            The strings 'major' and 'minor' can also be used to reference the major
            and minor axis.

            tag : str, optional -> Probe tag to be add a DataFrame column title.
        probe_units : str, option
            Units for probe orientation.
            Default is "rad".
```

With a probe class we would be able to do something like:

```python
probe_la = rs.Probe(15, Q_(45, 'deg'))
```

Another possibility would be to add filter (1x etc.) to the probe.

karenperezz:
Closes #827


probeclass.py
```diff
@@ -0,0 +1,13 @@
+class Probe():
+    @check_units
+    def __init__(self, node, angle, tag=None):
```
raphaeltimbo, line 3: Here we should have a docstring explaining each argument.

probeclass.py
```diff
@@ -0,0 +1,13 @@
...
+        self.node = node
+        self.angle = angle
+        if tag is not None:
+            self.tag = tag
+            
+    def probe_result(self):
+        if self.tag is not None:
+            return [(self.node, self.angle, self.tag)]
+        else:
+            return [(self.node, self.angle)]
```
raphaeltimbo, line 13: I did not understand how this method would be used in the code.
Could you please explain in more detail in the PR?

raphaeltimbo (changes requested):
Hi @karenperezz !
I have left some comments in the code.
Additionally, we should have some tests with this class.
The idea would be duplicate a test where we have the current use of a tuple to represent the probe, and replace this tuple with an object instantiated from the Probe class.

The name of the file could also be just 'probe.py' instead of probeclass.py.

ross-bott:
Hi there!
less than 45 days;

marcusfilipesr:
@raphaeltimbo, this PR is almost finished. It is lacking the tests and a few modifications that you asked for, @karenperezz is about to implement that.

marcusfilipesr (approved):

marcusfilipesr:
@raphaeltimbo, do you think we can merge this PR, after forcing update with `master`.

raphaeltimbo (approved):

PR #930, opened 2022-11-22
Fix lund method

elyqg:
Fix lund method in cylindrical bearing


elyqg:
Hello @raphaeltimbo. Necessary corrections to Lund's coefficients method are made. If there's anything else to fix just let me know.

Issue #573, opened 2020-06-01
Synchronous UCS Map


atbrandao:
ROSS currently considers no gyroscopic effects for the `plot_ucs()` method, as it calls `run_modal(speed=0)`.

The UCS Map should considers synchronous eigenvalues. That is to say that each plotted natural frequency considers the gyroscopic effects of the rotor at for `speed = frequency`.

This can be achieved by iterating eigenvalues, however this strategy would be very time consuming.
XLTRC2 uses a transformation of the M matrix that constrains the rotor speed to the eigenvalue frequency, simplifying this calculation. 
I extracted the paragraph below from XLTRC2 Help File:

> An undamped critical speed analysis is a special type of eigenvalue calculation. All bearings are reduced to simple undamped axisymmetric springs such that Kxx=Kyy and Kxy=Kyx=0 (see Brg's Worksheet). Also, **the rotor transverse inertia, It, is replaced with the transverse minus the polar inertia (It-Ip)**. This last modification, in effect, **constrains the rotor rotational speed to equal the eigenvalue frequency**. This means that computed eigenvalues are automatically synchronous (i.e., whirl frequency = rotor speed). Recall that eigenvalue frequencies normally vary with rotor speed because of gyroscopics. At particular values of rotor speed where the frequency of a forward whirling eigenvalue matches exactly the speed, the eigenvalue is termed a synchronous critical speed. This special command computes such eigenvalues directly, without the need to iterate the rotor speed.

Would it be possible to implement this transformation on ROSS?

raphaeltimbo:
I will start to implement this.
Some other points that I have noticed that maybe we will need to change:
- The undamped critical speed analysis is not a `run_` method, and I think this goes against our design. Maybe we should change this?;
- Should we change the name `ucs` to `undamped_critical_speed_analysis`? It has the con of being a long name, but is more helpful for those that are not used to use the name `ucs`. What do you guys think?;
- Another point is that the user does not have an option of choosing how many modes he can evaluate when calling the function. This goes back to a design that we used in the beginning of the project where some of this arguments would be defined as rotor attributes. Take for example a case where the user wants to determine the natural frequencies:
```python
rotor.n_eigen = 8    # define number of modes
rotor.speed = 1000   # define the speed and run eig()
rotor.wn             # get the frequencies
```

Now we have something like:
```python
rotor.n_eigen = 8
modal = rotor.run_modal(speed=1000)
modal.wn
```
I think we should evaluate if this makes more sense:
```python
[...]

PR #931, opened 2022-11-25
Dev sync ucs map

fernandarossi:
Now, the eigenvalues for the synchronous case are computed directly without the need to iterate the rotor speed.


raphaeltimbo:
Hi @fernandarossi !
This is looking good. 
Just to be sure that everything was giving the same results as the old code I added a test for the synchronous case and the expected results that would come from the old code.
The first thing I noticed, but can't exactly point the cause, is that the intersection points (where the support stiffness cross the natural frequency lines) are not the same as in the old cold.

To have a better understand of what was happening I ran the test and plotted the results.

This is the result with the old code:
image

And here with the new code:

image

If you want to reproduce this here is the code:

```python
import ross as rs

steel = rs.materials.steel

#  Rotor with damping
#  Rotor with 6 shaft elements, 2 disks and 2 bearings with frequency dependent coefficients
i_d = 0
o_d = 0.05
n = 6
L = [0.25 for _ in range(n)]

shaft_elem = [
    rs.ShaftElement(
        l,
        i_d,
        o_d,
        material=steel,
        shear_effects=True,
        rotary_inertia=True,
        gyroscopic=True,
    )
    for l in L
]

disk0 = rs.DiskElement.from_geometry(2, steel, 0.07, 0.05, 0.28)
disk1 = rs.DiskElement.from_geometry(4, steel, 0.07, 0.05, 0.35)

stfx = [1e7, 1.5e7]
stfy = [1e7, 1.5e7]
c = [1e3, 1.5e3]
frequency = [50, 5000]
bearing0 = rs.BearingElement(0, kxx=stfx, kyy=stfy, cxx=c, cyy=c, frequency=frequency)
bearing1 = rs.BearingElement(6, kxx=stfx, kyy=stfy, cxx=c, cyy=c, frequency=frequency)

rotor8 = rs.Rotor(shaft_elem, [disk0, disk1], [bearing0, bearing1])

ucs_results = rotor8.run_ucs(synchronous=True)
ucs_results.plot()
```

fernandarossi:
Hi @raphaeltimbo. If you run the old and the new code and compare the intersection points, you will observe that:

- The first intersection point is the same in both codes;
- The second, third, and fourth intersection points are different.

This occurs because when you run the old code, the speed is set to be equal to the first forward frequency. This is valid only for finding the critical speeds corresponding to the first mode. To find the critical speeds corresponding to the second mode, the speed should be set equal to the second forward frequency, and so on. If you do that, the other intersection points will give the same value.

ross/disk_element.py
```diff
@@ -89,11 +89,21 @@ def __eq__(self, other):
         True
         """
         false_number = 0
-        for i in self.__dict__:
-            try:
-                self.__dict__[i] == other.__dict__[i]
-            except:
-                false_number += 1
+        if other.__class__.__name__ == "DiskElement":
```
marcusfilipesr, line 92: Hi @fernandarossi, your lastest commit tests just failed, and I think the reason beghind is this line.

Since you are only comparing with the `DiskElement`, when building the `DiskElement6DoF` class it will go directly to the `else` statement, hence returning `False`. A possible solution would be:

```python
if "DiskElement" in other.__class__.__name__:
```

ross-bott:
Hi there!
less than 45 days;

ViniciusTxc3:
This is a new result with @fernandarossi  changes.

image

@raphaeltimbo This pr ends the issue #573

raphaeltimbo (approved):

PR #957, opened 2023-03-15
Mass Terms added to Bearings and Seals

### LINKED ISSUE #936: Add mass terms to bearing and seal elements
In pump seals mass terms can have a significant effect, especially for long seals where the direct mass coefficient can be quite large.
We should have the option to add the mass terms in the generic 
bearing and seal classes (`BearingElement` and `SealElement`).

marcusfilipesr:
This PR resolves #936


marcusfilipesr:
Hi @raphaeltimbo, I've just made the implementations. Is this what you had in mind?

raphaeltimbo:
Hi @marcusfilipesr! 
This is what I was thinking, yes.
But after looking at the tests (I've also added some more tests), I actually started to think about other points.
For example:
 - Should we allow the mass matrix to have a dependency on the frequency? 
 If yes, this will require additional changes to the code. Since we would also have to add this to the rotor mass matrix and the call to assemble the state space matrix (A).
 - Is it ok if we treat mxy and myx as always 0? And mxx and myy as always being equal? This way the user could have only a `m` argument and the interface would be more simple.

Since I do not have much experience with pump rotordynamics I will have to check what is the standard on those subjects. If you could also check on that before we proceed with the final implementation.

marcusfilipesr:
@raphaeltimbo, during the implementation I also had these thoughts.

Unfortunately, I too do not have experience with pump rotordynamics, but I'm going to search for information regarding the points that you raised.

marcusfilipesr:
Hi @raphaeltimbo, I was writing the code as we talked yesterday, and I came to this point. Below is the global mass method `M`  considering the mass in the bearings depending on the frequency.

```python
def M(self, frequency):
        """Mass matrix for an instance of a rotor.

        Returns
        -------
        M0 : np.ndarray
            Mass matrix for the rotor.

        Examples
        --------
        >>> rotor = rotor_example()
        >>> rotor.M()[:4, :4]
        array([[ 1.42050794,  0.        ,  0.        ,  0.04931719],
               [ 0.        ,  1.42050794, -0.04931719,  0.        ],
               [ 0.        , -0.04931719,  0.00231392,  0.        ],
               [ 0.04931719,  0.        ,  0.        ,  0.00231392]])
        """
        M0 = np.zeros((self.ndof, self.ndof))

        for elm in self.elements:
            dofs = list(elm.dof_global_index.values())
            try:
                M0[np.ix_(dofs, dofs)] += elm.M(frequency)
            except TypeError:
                M0[np.ix_(dofs, dofs)] += elm.M()
```

The reason I'm asking this, is because it will change quite a bit the way ROSS used to calculate the mass matrix. What do you think? I don't know if there is a clever way of writing this.

PR #958, opened 2023-03-16
Material density unit

### LINKED ISSUE #955: Fix units for material density
In the docstring for `ross.Material` the density `rho` is defined with units of `N/m**3` but in `Material.__str__` the units are `kg/m**3`. Based on the examples I've run, I believe the units for `rho` are `kg/m**3`. Also, the units should be written as `kg/m^3` instead of `kg/m**3`.

marcusfilipesr:
This PR resolves #955

### DIFF
--- a/ross/materials.py
+++ b/ross/materials.py
@@ -31,22 +31,25 @@ class Material:
     name : str
         Material name.
     rho : float, pint.Quantity
-        Density (N/m**3).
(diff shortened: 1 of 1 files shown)


raphaeltimbo (approved):
This is great. Thank you @marcusfilipesr!

PR #966, opened 2023-04-20
Plot modal shape 2D

### LINKED ISSUE #961: Orientation is not available to plot_2d
When running modal analysis and plotting the results we use:

```python
modal = rotor.run_modal(...)
modal.plot_mode_2d(...)
```
In `plot_mode_2d()` we do not have an option to pass `orientation='y'` to the method, although that is available in the `Shape.plot_mode_2d()` method. 

We should add the `orientation` argument to the `plot_mode_2d` method and add this to the docstring.

marcusfilipesr:
This PR closes #961


raphaeltimbo (changes requested):
Hi, @marcusfilipesr.
I would like to ask you for more 2 modifications here:
 - Add argument in the docstrings;
 - Add tests with each case: orientation='major' / ='x' and ='y'.

marcusfilipesr:
@raphaeltimbo have you faced something similar to this. When the test are running together, the one that I've just implemented fails. 

image

But when running alone, it passes.

image

raphaeltimbo:
> @raphaeltimbo have you faced something similar to this. When the test are running together, the one that I've just implemented fails.

I have faced something similar when there was no isolation, and one test would modify some object that was used by another test.

marcusfilipesr:
> > @raphaeltimbo have you faced something similar to this. When the test are running together, the one that I've just implemented fails.
> 
> I have faced something similar when there was no isolation, and one test would modify some object that was used by another test.

I think I've solved the problem.

raphaeltimbo (approved):

PR #972, opened 2023-05-02
Shape Normalization

### LINKED ISSUE #964: Change how we normalize the vectors in Shape objects
`Shape` objects have the option to normalize the vector which is passed in instantiation. This is used when creating shape objects for the mode shapes.
It would make sense that the max major axis in each normalized mode shape would be equal to one, but that is not the case for our current code.

marcusfilipesr:
This PR resolves #964


marcusfilipesr:
@raphaeltimbo, I was thinking, maybe if we change what is being compared in the tests, for example:

Instead of comparing orbits, Prof. Aldemir suggested of creating an interpolation between shape points, and comparing the coefficients of this polynom, disregarding the $a_{0}$.

PR #988, opened 2023-06-07
Smoother Amplitude Plots

### LINKED ISSUE #985: Plots for magnitude are not smooth
As we can see in the figure:

image

Magnitude plots for forced response analysis are not smooth.
Maybe we could use some interpolation process to make it look better.

marcusfilipesr:
This PR resolves #985


ross-bott:
Hi there!
less than 45 days;

Issue #1022, opened 2023-12-06
Error related to Kst matrix in the definition of the state space matrix A


jguarato:
![Problem_def_A](https://github.com/petrobras/ross/assets/82293939/16a161ca-14e2-4b22-8ab4-f9961d518cf4)

In the equation of motion of the rotor system, the stiffness matrix that results from the transient motion Kst should be multiplied by the acceleration [[Reference]](https://www.tandfonline.com/doi/full/10.1080/17415977.2015.1047364). However, in the definition of the state space matrix A in ROSS, Kst is multiplied by the rotor speed, as can be seen in the figure above.

jguarato:
There is also another stiffness matrix associated with the transient motion that is not yet considered in the model. As Kst is related to the shaft, the Kdt is related to the disk.

According to Lalanne e Ferraris (1998), this matrix can be obtained by:

$$
K_{DT} = \begin{bmatrix}
0 &0  &0    &0  \\
0 &0  &0    &0  \\
0 &0  &0    &0  \\
0 &0  &I_p  &0  \\
\end{bmatrix}
$$

ross-bott:
Hi there!
I have marked this issue as stale because it has not had activity for 45 days.
Consider the following options:
- If the issue refers to a large task, break it in smaller issues that can be solved in
less than 45 days;
- Label the issue as `wontfix` or `wontfix for now` and close it.

PR #1028, opened 2024-01-17
Fix use of Kst matrix

jguarato:
This PR resolve issue #1022 .

The Kdt matrix was also created for the disk element. However, this matrix is applied in time-dependent analyzes when there is speed variation. As the Kst matrix, it must be multiplied by the angular acceleration.

### DIFF
--- a/ross/disk_element.py
+++ b/ross/disk_element.py
@@ -638,6 +638,42 @@ class DiskElement6DoF(DiskElement):
         # fmt: on
         return K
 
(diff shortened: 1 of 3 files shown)


jguarato:
The Kdt matrix was also created for the disk element. However, this matrix is applied in time-dependent analyzes when there is speed variation. As the Kst matrix, it must be multiplied by the angular acceleration.

PR #1031, opened 2024-02-08
Fix disk K matrix

jguarato:
The K matrix for the disk elements should be zero. In the 6 dof model, there was confusion between the K matrix and the Kdt matrix of disk elements. This PR resolves this misunderstanding. Additionally, it proposes the incorporation of the Kdt matrix into the rotor Ksdt matrix, which represents the stiffness matrix related to transient motion of the shaft and discs (formerly Kst).


jguarato:
Hi @raphaeltimbo, 

It looks like the new version of _Black_ (24.1) is making the tests fail. I had version 23.10 installed, and I had no problems with it. However, when updating to the lastest version, the check with _Black_ reports several files that need to be formatted.

PR #1032, opened 2024-03-01
Add Newmark integrator in time response method

### LINKED ISSUE #1021: Incorporate time integration to account for speed variations
This issue was created in order to document and detail the task of implementing the numerical integrator that was assigned to me. This task may resolve other issues: #429, #884, #1001.

---

Currently, it is not possible to solve the transient model with the rotor speed varying over time in ROSS, nor to incorporate other information into the model that depends on time. One solution would be to use a numerical integrator to solve the following equation of motion for a rotor:

$$
M\ \ddot{q}+[C+ G\ \Omega(t)]\ \dot{q} + [K + K_{st}\ \dot{\Omega}(t)]\ {q} = F
$$

The function [`solve_ivp()`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.solve_ivp.html#scipy.integrate.solve_ivp) available in SciPy **integrate** library can be applied in this case. This function numerically integrates a set of first-order vector ordinary differential equations (ODEs):

$$
\frac{d\mathbf{y}}{dt} = \mathrm{f}(t,\ \mathbf{y})
$$

given the initial conditions $\mathbf{y}(0)=y_0$, and it also offers 6 different integration methods. Then, it would be possible to choose a more appropriate method for each type of problem, allowing to solve, for example, non-stiff problems using explicit methods (“RK23”, “RK45”, “DOP385”), and also stiff problems using implicit methods (“Radau”, “BDF”).

A higher-order ODE can be reduced to a first-order differential equation by introducing intermediate derivatives into the $\mathbf{y}$ vector. So, this implies that:

$$
\mathbf{y} = [q, \dot{q}]
$$

and accordingly $\mathrm{f}(t,\ \mathbf{y})$ must return:

$$
\frac{d\mathbf{y}}{dt} = [\dot{q}, \ddot{q}]
$$

where the second-order derivative could be calculated from the equation of motion by inverting the $M$ matrix as:

$$
\ddot{q} = M^{-1} [F - [C+ G\ \Omega(t)]\ \dot{q} - [K + K_{st}\ \dot{\Omega}(t)]\ {q} ]
$$

The values of rotational speed $\Omega(t)$ and acceleration $\dot{\Omega}(t)$ varying over time can be stored in arrays. Such arrays must be compatible with the array of times in which the computed solution will be stored.

Consequently, the integration can be done by calling:

```python
scipy.integrate.solve_ivp(fun, t_span, y0, method, t_eval, args)
```

and passing the callable function that will be integrated `fun(t, y)`, the interval of integration `t_span=[t0, tf]` and the initial state `y0`. It is optional, but it could be important to pass the integration method, the vector of times in which the computed solution will be stored `t_eval`, and a tuple of additional arguments that could be passed to `fun(t, y, *args)`.

At first, I defined the function `fun(t, y, *args)` to be integrated as follows:

```python
def _equation_of_motion(self, t, y, t0, dt, speed, accel, inv_M, C, G, K, Kst, F):
	
  size = int(len(y) / 2)
  disp_resp = y[:size]
  velc_resp = y[size:]

  i = round((t - t0) / dt)
  C_aux = C + G * speed[i]
  K_aux = K + Kst * accel[i]
  F_total = F[:, i]

  accl_resp = inv_M @ (F_total - (C_aux @ velc_resp) - (K_aux @ disp_resp))

  dydt = np.zeros(2 * size)
  dydt[:size] = velc_resp
  dydt[size:] = accl_resp

  return dydt
```

Here, I would like to highlight two points:

- Accessing the rotational speed `speed[i]` at a specific time `t` can be done by a conversion based on the initial time `t0` and time step `dt` to obtain the respective array index.
- Passing the model matrices as arguments makes the function generic, so it could be used to solve the model in both modal and physical spaces. However, this may change later if needed.

When choosing to solve in the modal space, there is a possibility to reduce model based on the selected number of modes. A reduced model implies a considerable reduction in computational cost. Another option, suggested by Prof. Aldemir, to obtain the reduced model would be to pass the  frequency range of interest.

jguarato:
Resolves #1021

---

Hi @raphaeltimbo,

I created this pull request to report what I have been doing regarding the numerical integrator. 
If there's anything I need to change, feel free to tell me.
I have some questions related to some of the implementations, and I will describe them better in the comments below.


jguarato:
## Description

The Newmark integrator was implemented to take into account the speed variation in the rotor time response method. It can be used by passing an argument, `integrator="newmark"`, to `run_time_response()` or passing `speed` as an array of values with the same length as `t` (time). The source of the new integrator is now available in `utils.py`.

The pseudo-modal method was also implemented to be considered in numerical integration, and it is only activated if the `num_modes` (number of modes) argument is passed to `run_time_response()`. In this method, the model is solved in the modal domain and can be reduced according to the first $n$ chosen modes. If `num_modes` is left out, the direct method is considered and the model is solved in the physical domain without reduction.

## Doubts

I am not sure if what I did on lines 891- 930 of `utils.py` is correct. To reduce the computational time, I tried to separate the bearing matrices from the other elements, and update them with the respective speed of the integration time step. This implementation was made for the cases where the bearing coefficients vary with speed, and the speed varies over time.

What do you think?

## Tests

Some tests were carried out to evaluate the integrator response and are presented below.

jguarato:
### Test 1

Model (with proportional damping):
  • 36 elements
  • 2 disks
  • 2 bearings (coefficients do not vary with speed and there are no cross coefficients)
rotor_lmest2

Analysis in the transient regime:
  • Max speed = 840 rpm (87.96 rad/s)
  • Unbalance at nodes [12, 23]
      ◦ Magnitude = [0.00034, 0.00034] kg.m
      ◦ Phase = [70, 30] degrees
  • Probes at nodes [3, 34]
speed_lmest2

After assembling the `rotor`, the simulation was run by making the following call:

```python
response = rotor.run_time_response(speed, F, t, integrator="newmark")
```

The results were compared against the results obtained with the code of LMEst - UFU:
13dof - lmest
14dof - lmest
137dof - lmest
138dof - lmest

Then, the reduced model was assumed by the pseudo-modal method for the first 12 modes:

```python
response = rotor.run_time_response(speed, F, t, integrator="newmark", num_modes=12)
```

The results were compared against the results obtained with the direct method:
13dof
14dof
137dof
138dof

In this case, the pseudo-modal method seemed to present results very similar to the direct method. The time taken to run both simulations was 243 s for the direct method (physical domain) and 22 s for the pseudo-modal method with 12 modes (approximately 91% reduction in computational cost).

jguarato:
I added a new argument that can be passed to the integrator `add_to_RHS`:

```python
"""
add_to_RHS : callable, optional
    An optional function that computes and returns an additional array to be added to
    the right-hand side of the equation of motion. This function should take arguments
    corresponding to the current state of the rotor system, including the time step
    number, displacements, and velocities. It should return an array of the same length
    as the degrees of freedom of the rotor system (`rotor.ndof`). This function allows
    for the incorporation of supplementary terms or external effects in the rotor system
    dynamics beyond the specified force input during the time integration process.
"""

def my_callable_function(step, displacement, velocity):

    ...

    return ext_force

	
response = rotor.run_time_response(speed, F, t, integrator="newmark", add_to_RHS=my_callable_function)
```

raphaeltimbo (approved):

PR #1111, opened 2024-09-17
Add MultiRotor class

jguarato:
This pull request introduces the `MultiRotor` class, initially designed to enable the coupling of multiple rotors using gears. To support this functionality, the `GearElement` class is also being implemented.

It is important to note that the use of these new classes requires the rotor model to support **6 degrees of freedom**.

I will provide updates here as the development progresses.


ross-bott:
Hi there!
less than 45 days;

plevisse-naarea:
Hey,
I'm particularly interested in this functionality! Thanks for tackling it. Do you have visibility on when it could be pulled into the main code ? Are there specific issues where I could potentially give a hand ?

jguarato:
> Hey, I'm particularly interested in this functionality! Thanks for tackling it. Do you have visibility on when it could be pulled into the main code ? Are there specific issues where I could potentially give a hand ?

Hey, @plevisse-naarea!

The main features of this module have already been implemented. I am currently trying to validate the results obtained from ROSS, following this paper [[A general method to predict unbalance responses of geared rotor systems]](https://doi.org/10.1016/j.jsv.2016.06.031) as reference.

For the first case presented in the reference, I achieved results that closely align:
Node3_2
Node8_2

However, for the second case, I am noticing some discrepancies and have not yet been able to identify the cause of these differences:
response_90

jguarato:
Apparently, the problem was related to the Gear diameter values ​​that were not informed in the paper. After some tests, we managed to obtain the correct values ​​and arrived at the following results:

newplot(3)
newplot(2)

Furthermore, we were able to perform a validation with the Example 10.9.1 case presented in the book by Friswell et al. (2010):
newplot(3)
newplot(4)

PR #1185, opened 2025-07-18
Add free-free analysis option in run_freq_response

jguarato:
The `run_freq_response` method of the `Rotor` class now includes an option to treat the rotor system as free-free. To enable this, simply pass the argument `free_free=True`.

**Example:**
```python
frequencies = np.linspace(0, 1000, 101)

rotor.run_freq_response(
     speed_range=frequencies, # Array with the desired range of frequencies in rad/s
     free_free=True, 
)
```

### DIFF
--- a/ross/rotor_assembly.py
+++ b/ross/rotor_assembly.py
@@ -1665,6 +1665,9 @@ class Rotor(object):
         >>> speed = 100.0
         >>> H = rotor.transfer_matrix(speed=speed)
         """
+        if frequency is None:
+            frequency = speed
+
         lti = self._lti(speed=speed)
         B = lti.B
         C = lti.C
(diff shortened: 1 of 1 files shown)


mcarolalb:
Previously, when using run_freq_response under free-free boundary conditions, the resulting FRF was incorrect. While we expected peaks at 107, 401, and 775 Hz, the obtained values were 106, 111, 322, 513, and 681 Hz instead.

`speed_range = np.linspace(0, 5000, 300)

results_frf =  rotor_mancal.run_freq_response(speed_range=speed_range)
plot_frf = results_frf.plot(inp = dof_hammer, out = dof_obs, amplitude_units = "m/s**2/N", frequency_units="Hz")`

image

Now, by setting the argument `free_free=True`, we obtained the correct FRF values, as shown below.

`freq = np.linspace(0, 5000, 300)

results_frf =  rotor_mancal.run_freq_response(free_free=True, speed_range=freq)
plot_frf = results_frf.plot(inp = dof_hammer, out = dof_obs, amplitude_units = "m/s**2/N", frequency_units="Hz")
`

image

Issue #1214, opened 2025-09-25
Error in determining the unbalance response at the bearing supports position


Raimundovpn:
I was analyzing the following rotor:

<img width="1459" height="490" alt="Image" src="https://github.com/user-attachments/assets/fbe2cacb-0f64-45d7-b3d7-165955a96b40" />

To create the bearing supports, nodes are added to the rotor model. However, to determine the unbalance response at these nodes, the variable self.rotor.nodes_pos[node] is required. Since the bearing support nodes are not found in this variable, attempting to run the following code:

```
speed = Q_(np.arange(0, 3001, 5), "RPM").to('rad/s').m
n1 = 8
m1 = Q_(118672,'g.mm').to('kg.m').m 
p1 = Q_(0,'rad')
n2 = 16
m2 = Q_(118672,'g.mm').to('kg.m').m 
p2 = Q_(0,'rad')
nn = 4
damped_unb_resp_1st = rotor.run_unbalance_response(node=[n1, n2],
                                                   unbalance_magnitude=[m1, m2],
                                                   unbalance_phase= [p1, p2],  
                                                   frequency=speed)
```
```
damped_unb_resp_1st.plot_magnitude(
    probe=[#rs.Probe(node=nn, angle=Q_(0,'deg'),  tag="DE Probe x"),
           #rs.Probe(node=nn, angle=Q_(90,'deg'), tag="DE Probe y"),
            rs.Probe(node=31, angle="major", tag="NDE Probe major"),
            rs.Probe(node=32, angle="major", tag="DE Probe major"),
            # rs.Probe(node=31, angle="minor", tag="NDE Probe minor"),
            # rs.Probe(node=32, angle="minor", tag="DE Probe minor"),
            ],
    amplitude_units="mm/s pkpk",
    frequency_units="RPM",
)
```

Results in the following error:
```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Cell In[11], line 1
----> 1 damped_unb_resp_1st.plot_magnitude(
      2     probe=[#rs.Probe(node=nn, angle=Q_(0,'deg'),  tag="DE Probe x"),
      3            #rs.Probe(node=nn, angle=Q_(90,'deg'), tag="DE Probe y"),
      4             rs.Probe(node=31, angle="major", tag="NDE Probe major"),
      5             rs.Probe(node=32, angle="major", tag="DE Probe major"),
      6             # rs.Probe(node=31, angle="minor", tag="NDE Probe minor"),
[...]

Issue #1208, opened 2025-09-05
Divergence when using the run_unbalance_response function


Raimundovpn:
***Problem Description***
While attempting to reproduce the results from a report for a specific rotor, a significant divergence was observed between the unbalance response calculated using the standard ross method and the expected results.

Below is the reference response from the report:
<img width="861" height="702" alt="Image" src="https://github.com/user-attachments/assets/19128097-36c4-4007-a539-153f49a59370" />

To reproduce this result, the standard run_unbalance_response method was used with the following code:

```
import numpy as np
from ross.units import Q_

speed = Q_(np.arange(0, 3001, 5), "RPM").to('rad/s').m
n1 = 8
m1 = Q_(118672,'g.mm').to('kg.m').m
p1 = Q_(0,'rad')
n2 = 16
m2 = Q_(118672,'g.mm').to('kg.m').m
p2 = Q_(0,'rad')
nn = 4

damped_unb_resp_1st = rotor.run_unbalance_response( node=[n1, n2],  unbalance_magnitude=[m1, m2],   unbalance_phase=[p1, p2], frequency=speed)
```

The obtained result was as follows:

<img width="824" height="433" alt="Image" src="https://github.com/user-attachments/assets/43cc426a-4905-49f6-aac1-08d59753e76b" />

<img width="822" height="419" alt="Image" src="https://github.com/user-attachments/assets/ae9300d3-ee52-4604-8658-eccb9eb9d90c" />

*Solution Attempts and Analysis*

Initially, the discrepancy was attributed to potential errors in the rotor modeling. However, after a conversation with @jguarato, it was confirmed that she encountered the same issue. She suggested an alternative approach using an iterative run_time_response for each frequency and collecting the maximum amplitude. Her code, which successfully reproduces the correct response, is shown below:
```

import numpy as np
import plotly.graph_objects as go
from tqdm import tqdm
from ross.units import Q_

speed_range = Q_(np.arange(0, 3001, 10), "RPM").to('rad/s').m
t = np.arange(0, 6, 1e-4)

n1 = 8
m1 = Q_(118672,'g.mm').to('kg.m').m
p1 = Q_(0,'rad')
n2 = 16
m2 = Q_(118672,'g.mm').to('kg.m').m
p2 = Q_(0,'rad')
probe = 4

num_dof = rotor.number_dof

probe_x = []
probe_y = []

for speed_ in tqdm(speed_range, desc="Velocidade"):
[...]

Issue #1153, opened 2025-03-07
Inaccurate Unbalance Response for rotors with cross coupled journal bearings


PeterSinelnikov:
I was running some tests to compare unbalance response results with other commercial software that I've been using for years and have found that it produces results that are very close to the real-world unbalance responses.

ROSS library produced results that were very close to the commercial counterpart, but as soon as I introduce cross-coupled bearing coefficients Cxy and Cyx, the results are not what I would expect.

![Image](https://github.com/user-attachments/assets/969734e8-cbca-4c47-9462-6b9f74f6f510)

Compare to this output:

![Image](https://github.com/user-attachments/assets/6fec9588-9606-4379-8e84-a0d4ded010e5)

In both cases I use identical setup, the same rotor model, the same bearing coefficients, supports, point masses and so on.

Bearing coefficients I'm using are as follows:

![Image](https://github.com/user-attachments/assets/2276810a-852c-4a54-a9ea-e91b4a014746)

Could anybody comment on this? It fills like there's something how it accounts for Cxy and Cyx that throws it off.

jguarato:
Hi @PeterSinelnikov!

Could you share your script so we can take a closer look at the issue?

Based on the information and results you provided, we think this may be related to the damping of the system. Are you considering proportional damping in your model? By default in ROSS, if the values ​​for the `alpha` and `beta` coefficients are not specified when instantiating the `ShaftElement`, the proportional damping is set to zero.

We ran some tests to check this issue and obtained for the same model:

- Without proportional damping:
![Image](https://github.com/user-attachments/assets/4fff5349-ba0a-459e-b873-1ec929a49029)

- With proportional damping:
![Image](https://github.com/user-attachments/assets/0dfda6dd-a1cd-42a6-a29c-2154a3b8a687)

PeterSinelnikov:
@jguarato Thank you for taking the time to look into this. I appreciate it.

Here's my code:
```
import time
from pathlib import Path

import numpy as np

import ross as rs
from ross import Probe, Q_

shaft_file = Path("gen_rotor.xls")
shaft = rs.ShaftElement.from_table(
    file=shaft_file, sheet_type="Model", sheet_name="Model"
)
start_time = time.time()
bearing1 = rs.BearingElement(n=6, n_link=51, kxx=469218000, kxy=-152207000, kyx=-1328630000, kyy=2408770000,
                             cxx=1477630, cxy=-2479570, cyx=-2479570, cyy=12537000)
[...]

PR #1216, opened 2025-10-01
Frf

Raimundovpn:
This pull request introduces a new function for calculating the unbalance response. This was developed to address inconsistencies and divergences observed between the results generated by the existing run_unbalance_response function and those from other established tools.

These discrepancies have been documented in the following issues:

Divergence when using the run_unbalance_response function (#1208)

Error in determining the unbalance response at the bearing supports position (#1214)

Inaccurate Unbalance Response for rotors with cross-coupled journal bearings (#1153)

The new implementation demonstrates a clear convergence with results from technical reports and other software, ensuring more reliable and accurate simulations.


ross/results.py
```diff
@@ -3233,6 +3205,43 @@ def __init__(
             "[length] / [time] ** 2": ["m/s**2", "accl_resp"],
         }
 
+    def _calculate_amplitude(self, angle, ru_e, rv_e):
```
jguarato, line 3208: Change function name to `calculate_amplitude` (remove underline), and then update `test_orbit_calculate_amplitude` in  _ross/tests/test_results.py_.

jguarato (approved):

PR #1280, opened 2026-04-10
Rotor id summary

kiracofe8:
Addresses issue https://github.com/petrobras/ross/issues/1265

Two things I was not sure of that I would like feedback on:

1) When I ran the ruff formatter, it changed many lines of code that were completely unrelated to my changes.  E.g. in docs/user_guide/tutorial_part_5.ipynb  I can't tell if this means that previous contributors simply did not run ruff on their code, or if I have run it incorrectly.  This is my first time ever using ruff.  

2) I added several test cases.  My philosophy is you can't have too many test cases, but it does slow the process down a bit to run all the test cases. So if you want I can scale it back to just one basic test case.

### DIFF
--- a/ross/results.py
+++ b/ross/results.py
@@ -5290,13 +5290,14 @@ class SummaryResults(Results):
         The figure object with the tables plot.
     """
 
-    def __init__(self, df_shaft, df_disks, df_bearings, brg_forces, CG, Ip, tag):
+    def __init__(self, df_shaft, df_disks, df_bearings, brg_forces, CG, Ip, It,tag):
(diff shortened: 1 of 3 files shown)


raphaeltimbo:
Thanks for the contribution!

  1. Unrelated tutorial_part_5.ipynb changes

  Not your fault. ruff format --check was fixed in CI recently in #1277 (merged after this tutorial was added). So don't worry about that

  2. Tests

  We can keep the tests since the extra runtime I believe is negligible and the tests are useful.

Issue #1276, opened 2026-04-06
Add `ruff format --check` to CI — fails PRs when unformatted files are detected


gsabinoo:
## Background

This repository uses Ruff in two distinct modes:

- **`ruff check`** — enforces lint rules defined under `[tool.ruff.lint]` in `pyproject.toml`.
- **`ruff format`** — applies automatic formatting governed by `[tool.ruff.format]`.

These two modes are **not interchangeable**: a file can pass every lint rule and still differ from what the formatter would produce.

---

## What changed in CI

The test workflow now also runs:

```bash
ruff format --check ross
```

As a result, the pipeline validates not only lint compliance but also whether the code is formatted according to Ruff's style. If any file requires formatting-only changes, that step exits with a non-zero code and the job is marked as failed — even when `ruff check` reported nothing locally.

---

## How to fix it locally

Before opening or updating a PR, run both steps:

```bash
ruff check ross
ruff format ross
```

To validate without writing changes:

```bash
ruff format --check ross
```

> **Tip:** If the project includes a `pre-commit` configuration, the `ruff` and `ruff-format` hooks already cover both steps automatically.

---

## Purpose of this issue

This issue aims to document the observed behavior and help prevent confusion when the Ruff CI job fails due to pending formatting. It also clarifies that **a successful lint check does not necessarily indicate that the code is fully formatted according to Ruff’s standards**.

PR #1305, opened 2026-05-15
SummaryResults units conversions

kiracofe8:
Addresses Issue 1282 https://github.com/petrobras/ross/issues/1282


gsabinoo:
Hey @kiracofe8! We recently added `ruff format --check` to our CI pipeline (see #1276), so I took the liberty of running `ruff format ross` on this branch and committing the formatting changes. No logic was touched — purely style adjustments to keep CI green. Thanks for your contribution!

raphaeltimbo:
Thanks a lot for this contribution, @kiracofe8. Adding unit options to the summary tables is a nice usability improvement and rounds out the units support we already have in the other results plots.

I pushed a small follow-up commit (0cfd69c8) to align the change with a few ROSS conventions - nothing about the behavior, just style/quality:

- **Docstring**: reformatted to NumPy style (`length_units : str, optional` with a description and `Default is '...'.` for each param), and moved the "no consistency check" caveat into a proper `Notes` section.
- **f-strings**: replaced the `"text (" + units + ")"` header concatenations with f-strings (e.g. `f"Length ({length_units})"`), which is the idiom used elsewhere in `results.py`.
- **Inertia units**: display them with a superscript (`mm²*g`) instead of `mm^2*g` (the pint conversion string still uses `**2`).
- **Test**: extended `test_summary_results` to actually validate the conversions (e.g. `Length (mm) == Length (m) * 1000`) and to cover the default-units path, in addition to the smoke checks.

I went with the three-independent-units approach you chose (rather than a pint unit *system*) — it's simpler and the `Notes` caveat makes the trade-off explicit.

All tests pass locally and ruff is clean. Merging now - thanks again.

PR #1328, opened 2026-07-24
Add 3D pad temperature plot and clarify bearing temperature plot names

raphaeltimbo:
## Summary

Bearing temperature plotting gets two clearly-named 3D views instead of one ambiguous one, and the through-pad view is implemented for the tilting pad.

`plot_temperature_3d` is renamed to **`plot_film_temperature_3d`**, so the name says which temperature is drawn. The old name stays as a deprecated alias that forwards its arguments and warns.

A new **`plot_pad_temperature_3d`** draws the pads as real geometry colored by the solid pad temperature.

The two are complementary, and the reason for two names is that they resolve different dimensions:

| Method | Draws | Resolves |
|---|---|---|
| `plot_film_temperature_3d` | height-map surface of the oil film field | circumferential × **axial** |
| `plot_pad_temperature_3d` | mesh of the pad geometry, colored by temperature | circumferential × **radial** |

Neither is a superset of the other, so a single shared name would have promised a dimension that is not computed.

## `plot_pad_temperature_3d`

Implemented for `TiltingPadResults`, which already solves a radially resolved pad conduction field (`T_pad`, shaped `(n_pad, ntheta_pad, nr_pad)`). Each pad is drawn at its true angular position, spanning the pad thickness from the babbitt surface out to the pad back, with unit selection and an optional dashed babbitt-surface outline.

It is present on every result class so the plotting API is uniform; models that compute no solid pad field raise `NotImplementedError` naming the class. `TiltingPadResults` is currently the only implementation.

Because the solver does not resolve the solid temperature along the bearing axis, the axial extrusion is geometric — both axial faces carry the same temperature. This is stated in the docstring, since it is easy to misread from the plot.

## Other changes

- `plot_results()` now returns the key `"film_temperature_3d"` in place of `"temperature_3d"`, matching the renamed method.
- `TiltingPadResults` receives `offset`, needed to place each pad at its leading edge (`pivot_angle - offset * pad_arc`). It falls back to a centered pivot when absent, so manually built results objects still work.
- `plot_pad_temperature_3d` rejects a non-zero `freq_index`: the tilting pad solver keeps a single `T_pad` for the last solved speed, so indexing other frequencies would silently return the wrong field.
- Docs: new section 2.13 in `tutorial_bearings_part_2` showing both plots, and a correction to the documented `plot_results()` key names, which listed `'temperature_3D'` and `'pressure_3D'` with a capital D that never matched the real keys.

## Verification

Checked against a real solve (5-pad tilting pad, Ø101.6 mm, ISOVG32, 3000 RPM):

- node count `4500 = 5 pads × 30 θ × 15 radial × 2`
- intensity range equals the `T_pad` range exactly
- radial span equals `pad_radius` → `pad_radius + pad_thickness` exactly
- each pad spans its own pivot ±30°
- `mm` and `degF` conversions exact

All bearing test suites pass (145 tests, including 8 new), ruff and doctests clean, and the tutorial notebook executes without errors.

## Backwards compatibility

The only breaking change is the `plot_results()` dictionary key. The method rename is covered by the deprecated alias.

### DIFF
--- a/ross/bearings/bearing_results.py
+++ b/ross/bearings/bearing_results.py
@@ -146,9 +146,7 @@ class BearingResults(ABC):
             "pressure_2d": self.plot_pressure_2d(freq_index=freq_index),
             "pressure_3d": self.plot_pressure_3d(freq_index=freq_index),
             "temperature_2d": self.plot_temperature_2d(freq_index=freq_index),
-            "film_temperature_3d": self.plot_film_temperature_3d(
-                freq_index=freq_index
-            ),
+            "film_temperature_3d": self.plot_film_temperature_3d(freq_index=freq_index),
         }
 
         if show_plots:
--- a/ross/bearings/bearing_results.py
+++ b/ross/bearings/bearing_results.py
@@ -1,3 +1,4 @@
+import warnings
 from abc import ABC, abstractmethod
 
 import numpy as np
@@ -7,6 +8,7 @@ from prettytable import PrettyTable
 from scipy.interpolate import griddata
 
 from ross.plotly_theme import tableau_colors
+from ross.units import Q_
 
 __all__ = [
     "BearingResults",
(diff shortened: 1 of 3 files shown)


raphaeltimbo:
Closing: #1332 replaced the bearing engine and rewrote the result classes this PR was built on, so the implementation here no longer applies. Its content landed in two places on the new stack:

- The naming clarification — `plot_temperature_3d` → `plot_film_temperature_3d` with a deprecated alias, and the `plot_results()` key — was folded directly into #1332 (commit 77f3c4ee).
- `plot_pad_temperature_3d` was reimplemented on the new solver in #1339, built on the energy equation's solid pad field. The new version lifts two limitations of this one: it works per-frequency (the old engine kept a single `T_pad` for the last solved speed) and for any conducting-pad bearing, not just tilting pads.

PR #1347, opened 2026-08-13
Use loky executors for coefficient solver pools

raphaeltimbo:
## Summary
The fork-context `multiprocessing.Pool`s in `LabyrinthSeal`, `HolePatternSeal` and `FluidFilmBearing._solve_all` crash inside Jupyter kernels: forked workers inherit the kernel's ZMQ streams and die with `ZMQError: Socket operation on non-socket` (reproduced on pyzmq 25 and 27).

All three pools now use [loky](https://github.com/joblib/loky)'s `get_reusable_executor` (the executor behind joblib/scikit-learn), added as a dependency. Loky starts fresh spawn-based workers, so nothing is forked and no kernel sockets are inherited — and unlike the stdlib `spawn` context, its workers do **not** re-import `__main__`. A plain `multiprocessing.get_context("spawn")` pool was considered first, but an unguarded script hangs forever with it: each spawned worker re-imports the script, crashes during bootstrap, and the pool silently respawns it (~32k lines of repeated tracebacks in 20 s, never exits). With loky that failure mode does not exist.

**Behavior note:** no `if __name__ == "__main__":` guard is needed on any platform — plain scripts that construct these elements at module level keep working on Linux, Windows and macOS, and Jupyter kernels are unaffected. Worker count is capped at the number of frequency cases, and the executor is reused across calls, amortizing worker startup.

## Test plan
- labyrinth/holepattern/fluid-film element suites: 43 passed
- fluid-film solver/literature/docstrings + `ross/stochastic`: 91 passed, 3 skipped (REFPROP-gated)
- Unguarded plain script constructing `LabyrinthSeal` (5 freqs) at module level: completes normally (would hang forever with a stdlib spawn pool)
- Notebook executed through nbclient/ipykernel exercising the pooled path (LabyrinthSeal 5 freqs): completes with no ZMQError, results identical to the script run
- ruff check and format clean

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,3 +10,4 @@ numba
 ccp-performance
 prettytable
 control
+loky>=3.0
--- a/ross/bearings/fluid_film_bearing.py
+++ b/ross/bearings/fluid_film_bearing.py
@@ -13,10 +13,10 @@ tilting pad, ...) subclass this and only translate their user-facing
 geometry into the per-pad arrays this class consumes.
 """
 
-import multiprocessing
 import time
 
 import numpy as np
+from loky import get_reusable_executor
 
 from ross.bearing_seal_element import BearingElement
 from ross.bearings.bearing_results import FluidFilmBearingResults
@@ -698,8 +698,13 @@ class FluidFilmBearing(BearingElement):
         """
         per_case = [self._engine_inputs(f) for f in self.frequency_range]
         if num_processes is not None and num_processes > 1:
-            with multiprocessing.Pool(num_processes) as pool:
-                return pool.map(_solve_case, per_case)
+            # loky starts fresh workers (no fork, so no inherited sockets
+            # crashing Jupyter kernels) and does not re-import __main__,
+            # so scripts need no if __name__ == "__main__" guard.
+            executor = get_reusable_executor(
+                max_workers=min(num_processes, len(per_case))
+            )
+            return list(executor.map(_solve_case, per_case))
         return [_solve_case(inputs) for inputs in per_case]
 
     def coefficients(self, frequency):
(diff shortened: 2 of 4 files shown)


raphaeltimbo:
Closing this PR — further investigation showed its premise was a misdiagnosis, and the fork pools on `main` are fine.

**What the "ZMQError in Jupyter" actually was:** teardown noise from a notebook-execution timeout, not a fork/ZMQ bug. The docs runner used `NotebookClient(timeout=600)`; the failing cell (`rs.TiltingPad`, which solves *sequentially* — `num_processes=None`) legitimately takes longer than that. When nbclient timed out, it shut the kernel down mid-execute, and the kernel logged `ZMQError: Socket operation on non-socket` while trying to send its reply on the already-closed shell socket. The signature reproduces exactly with a bare `time.sleep(60)` cell and a 5 s timeout — no multiprocessing involved.

Additional evidence:

- Fork pools run clean inside real ipykernels (tested 6.31, 7.2 and 7.3, pyzmq 27.1), including the actual `LabyrinthSeal` pooled solve and a stress test with workers writing hundreds of prints/warnings into the kernel streams.
- Two of the three "crashing" notebooks (`example_31`, `example_32`) only use `BearingElement`/`Rotor`/`ShaftElement` — they never touch the pools.
- The run that appeared to confirm the spawn fix had also raised the timeout to 3600 s (`tutorial_bearings_part_2` takes ~3100 s), which is what actually made it pass.

The spawn/loky change also carried a real cost: spawned workers re-import ross (~13 s cold start on a 4-core machine vs ~0.4 s for the fork pool), penalizing plain Linux scripts to fix a bug that doesn't exist.

Possibly worth revisiting separately: Python 3.14 changes the default start method on Linux from fork to forkserver, so the pools' behavior there deserves a look when we get to 3.14 support — but driven by an actual failure, not this ghost.

PR #1395, opened 2026-09-15
Move all workflows to the Node 24 majors of the GitHub actions

raphaeltimbo:
## Summary

Follow-up to #1392. The first CI run of that PR, at `checkout@v4` / `setup-python@v5`, still produced the runner annotation:

```
Node.js 20 is deprecated. The following actions target Node.js 20 but are being
forced to run on Node.js 24: actions/checkout@v4, actions/setup-python@v5.
```

The interface workflow gets the same annotation for `setup-node@v4` and `upload-artifact@v4`, and the publish workflow merged in #1391 for `checkout@v4` / `setup-python@v5`. The current majors that declare `runs.using: node24` are `checkout@v7`, `setup-python@v7`, `setup-node@v7` and `upload-artifact@v7` (checked in each action's `action.yml` at its latest release).

- `test.yml`, `publish-to-pypi.yml`, `interface.yml` and its mirrored copy `ross/interface/ci/interface.yml` move to those majors.
- `ross/interface/tests/test_packaging.py` pins the `upload-artifact` major in one assertion; updated.
- `softprops/action-gh-release@v2` and `pypa/gh-action-pypi-publish@release/v1` are unchanged (composite / not flagged).

## Verification

- `pytest ross/interface/tests/test_packaging.py`: 41 passed, both workflow copies identical.
- The annotation should be absent from this PR's runs.


raphaeltimbo:
Verified on this PR's first finished jobs (interface `check (ubuntu-latest, 3.12)` and `check (windows-latest, 3.12)`, which run checkout@v7, setup-python@v7 and setup-node@v7): zero annotations. The same jobs on the v4/v5 pins carried "Node.js 20 is deprecated ... forced to run on Node.js 24". The build matrix is still queued behind other runs.

PR #1409, opened 2026-09-15
Release candidate 3.0.0rc1

raphaeltimbo:
Step 3 of the release sequence in CONTRIBUTING.md: set `__version__ = "3.0.0rc1"` so the release path (PyPI trusted publishing, the Windows interface bundle attached to a GitHub pre-release, the Read the Docs build) can be exercised on a candidate before `v3.0.0` is tagged.

Also finishes the release notes for the PRs merged after #1403 completed them:

- #1407 (Tests on `maintenance/**`, `Release` on `v[0-9]*` tags) in *Packaging and Release Infrastructure*;
- #1405 and #1408 (DocSearch live) replace the "inactive until credentials are available" sentence in *Documentation*;
- #1402 (README links) and #1404 (overview tutorial `speed=`) as a new *Documentation* bullet.

`CITATION.cff` is deliberately left as is.

Verified: sdist and wheel built from a clean clone carry `Version: 3.0.0rc1`; the wheel installs with the `interface` extra into a fresh Python 3.12 venv and `ross-interface --version` and `--selftest` report `3.0.0rc1`; the release-notes and version tests pass; the notes parse with docutils.

After merge, from a clean clone of the merged commit:

```
git tag v3.0.0rc1
git push upstream v3.0.0rc1
```

then publish a GitHub pre-release from the tag.


raphaeltimbo:
The macOS and Windows failures are pre-existing on `main` (the run on 74e2b011 fails the same test on all five macOS and Windows jobs): `test_cylindrical_rejects_speed_without_equilibrium` from #1406 expects the `ValueError` at zero speed, but on those platforms the root finder returns one real root just below 1 for the quadruple root of `(x - 1)^4`, it passes the `0 < r < 1` filter, and the coefficient formulas divide by zero, so scipy raises "`y` must contain only finite values" instead. Linux returns four complex roots, which is why the test passed locally and on ubuntu.

Pushed a second commit that rejects `speed <= 0` explicitly before the root search, plus a sentence in the 3.0.0 notes. Verified locally by injecting the macOS root values: the intended `ValueError` is raised.

PR #1410, opened 2026-09-15
Release 3.0.0

raphaeltimbo:
Step 4 of the release sequence in CONTRIBUTING.md: `__version__ = "3.0.0"`. The release notes were completed in #1403 and #1409; `CITATION.cff` is deliberately left as is.

Opened while the 3.0.0rc1 candidate is still being verified so that its checks can queue in parallel. **Merge only after the candidate checks out**: PyPI install (done, `pip install --pre "ross-rotordynamics[interface]"` resolves to 3.0.0rc1 with plotly 7.1.0, modal analysis and `ross-interface --selftest` pass), the Windows zip on the pre-release (interface release run in progress) and the `en/v3.0.0rc1` docs build (version to be activated on Read the Docs).

After merge, from a clean clone of the merged commit:

```
git tag v3.0.0
git push upstream v3.0.0
```

then publish the GitHub release with the 3.0.0 notes as body, create `maintenance/3.0.x` from the tag and bump `main` to `3.1.0.dev0`.


raphaeltimbo:
The nine matrix jobs passed; the failure was in the extra "Installed package tests" job, on `test_run_amb_non_collocation`: it got `mode_indices` `[0, 1, 2, 3, 4, 7]` instead of the hard-coded `[0, 1, 2, 3, 6, 7]`, i.e. the two torsional modes came out at indices 5 and 6 on that runner. The same job passed on both rc branch runs with the same library tree, so the position of the torsional pair in the eigenvalue ordering is not stable across runs.

Pushed a second commit that asserts the partition (kept plus excluded indices equal the eight requested, exactly two excluded, both torsional) instead of fixed positions. Test-only change.

PR #1414, opened 2026-09-21
Accept speed range end points in run_campbell and the frequency responses

viniciuserrav:
### Summary

`Rotor.run_campbell`, `Rotor.run_unbalance_response` and `Rotor.run_freq_response` gain three optional keyword arguments — `speed_start`, `speed_stop` and `num_points` — so a speed sweep can be given by its end points instead of an explicit array. Both end points go through `@check_units`, so each one can be given in its own unit:

```python
rotor.run_campbell(speed_start=Q_(0, "RPM"), speed_stop=Q_(60, "Hz"))
rotor.run_unbalance_response(
    3, 1e-4, 0, speed_start=0, speed_stop=Q_(10000, "RPM"), num_points=500
)
```

### Motivation

Today the sweep has to be built by hand with `np.linspace`. A natural `(start, stop)` guess is accepted silently but runs the analysis at those two speeds only: `rotor.run_campbell((0, 400))` returns a two-point Campbell diagram.

### Default number of points

When `num_points` is not given, it is `ceil(k * sqrt(span))`, with the span in rad/s, clipped to [20, 2000]. The number of points grows with the range but more slowly than the range itself, so a narrow range (e.g. a zoom around a critical speed) gets a finer step than a wide one.

- `k = 20` for the frequency responses.
- `k = 4` for the Campbell diagram, whose curves are smooth and where each point is a full modal analysis. This gives 50–90 points over ~4000 RPM, which matches the ranges used in the documentation examples.

| Range | Responses (k = 20) | Campbell (k = 4) |
|---|---|---|
| 2800–3200 RPM | 130 points, 3.1 RPM step | 26 points |
| 0–3000 RPM | 355 points, 8.5 RPM step | 71 points |
| 0–10000 RPM | 648 points, 15.5 RPM step | 130 points |
| 0–30000 RPM | 1121 points, 26.8 RPM step | 225 points |

No uniform grid guarantees the peak amplitude of a lightly damped mode (ζ below ~1%). For that, pass `num_points` or narrow the range around the critical speed.

### Compatibility

- The new arguments are appended at the end of the signatures, so existing calls behave as before, including the 1000-point default of the responses when no range is given.
- `speed_range` of `run_campbell` becomes optional. Calling it with neither `speed_range` nor the end points raises `ValueError`.
- Giving `speed_range` together with the end points, only one end point, or `num_points` without the end points raises `ValueError`.
- Left unchanged on purpose:
  - `run_forced_response`, because its force array is tied to the sweep.
  - `run_clearance_analysis`, because its `speed_range` comes before required positional arguments.
  - `run_critical_speed`, because its `speed_range` already means a `(start, end)` filter.

### Changes

- `ross/utils.py`: `make_speed_range` (builds the array, with `@check_units`) and `resolve_speed_range` (validates the combination of arguments).
- `ross/rotor_assembly.py`: the three methods, with docstrings and doctest examples.
- `ross/tests/test_rotor_assembly.py`: tests for the point-count rule, unit handling, equivalence with an explicit array, and argument errors.

### Tests

- `pytest ross -n auto`: 907 passed, 9 skipped (896 before this change, plus 11 new).
- `ruff check` and `ruff format --check` pass on the changed files.

### Contributor License Agreement

- [x] I have read and agree to the [Individual Contributor License Agreement](https://github.com/petrobras/ross/blob/main/clas/individual_cla.md).


ViniciusTxc3:
Hi @viniciuserrav, thanks for putting this together.

I've been going through the change and wanted to raise a couple of points before it goes further.

On where this method should live: `speed_range` in `run_campbell`, `run_freq_response`, and `run_unbalance_response `already accepts any array-like (or `pint.Quantity`), so the endpoint-based ergonomics don't require new parameters on three separate public methods. I'd suggest keeping make_speed_range (or a more explicit name like `speed_range_from_endpoints`, to distinguish it clearly from the existing `make_speed_array`, which builds a time-domain speed ramp. The naming overlap could confuse newcomers) as a standalone function in utils.py, called explicitly by the user:

```python
from ross.utils import speed_range_from_endpoints

speed_range = speed_range_from_endpoints(Q_(0, "RPM"), Q_(3600, "RPM"))
camp = rotor.run_campbell(speed_range)
```

rather than folded into `resolve_speed_range `dispatch logic inside each `run_*` method. This avoids adding `speed_start`/`speed_stop`/`num_points `plus the mutual-exclusion validation to three signatures, while still giving users the same one-liner convenience. The `run_campbell`/`run_freq_response`/`run_unbalance_response `docstrings could simply point to it. E.g. a "See Also" note along the lines of "use ross.utils.speed_range_from_endpoints to build speed_range from its end points", so the pattern is discoverable without expanding the core method signatures. 
I believe this feature could even help us with issues that occurred in the past, such as with issue #985.

Do you see an advantage to embedding this directly in the `Rotor` methods that the standalone-utility approach wouldn't give us, maybe a use case in the interface app or elsewhere where resolving it inside Rotor matters? 
Happy to be convinced either way, just want to make sure the added parameters/branches on the core methods are earning their complexity. 
Thanks again for the work here!

viniciuserrav:
Hi @ViniciusTxc3, thanks for the careful review. I agree with the standalone approach and have pushed it.

**On the use case that would justify resolving it inside `Rotor`**: I went looking in the interface and found the opposite of a justification. It never passes end points into the `run_*` methods, it builds the array itself before calling them:

- `interface/services/analysis/campbell.py` and `clearance.py`: `np.linspace(spec["speed_min"], spec["speed_max"], spec["steps"])`
- `interface/services/analysis/freq_response.py` and `unbalance.py`: `np.linspace(spec["speed_min"], spec["speed_max"], 50)`

So the interface is a consumer of the standalone function, not of method parameters. Those two hardcoded `50`s are exactly the case the helper is meant for, since the number of points is the same whether the user asks for 0–500 RPM or 0–50000 RPM. I left them alone here to keep the diff to the point, but I am happy to switch the four call sites, in this PR or in a separate one.

**What the parameters were giving that the function does not:** each method had its own default point count, `20 * sqrt(span)` for the responses and `4 * sqrt(span)` for the Campbell diagram, whose curves are smooth and whose every point is a full modal analysis. That is now the `points_factor` argument, defaulting to 20, with the Campbell case documented in the function and in the `See Also`.

**What changed:**

- `ross.utils.speed_range_from_endpoints(speed_start, speed_stop, num_points=None, points_factor=20)`, with the name you suggested, to keep it clear of `make_speed_array`.
- `speed_start`, `speed_stop`, `num_points` and the mutual-exclusion validation are gone from `run_campbell`, `run_freq_response` and `run_unbalance_response`. Those methods are back to `main` except for a `See Also` block in each docstring.
- The tests moved to `ross/tests/test_utils.py`, including one that feeds the range into `run_campbell` and `run_unbalance_response`.

**On #985:** this helps only partly, and I would rather not claim it closes that issue. The rule here is uniform spacing whose point count grows with the range, so a narrow sweep around a critical speed does get a finer step, but it does not put extra samples near the natural frequencies, which is what @raphaeltimbo suggested in that thread. Related to it: `_run_freq_response` still documents `cluster_points`, `num_modes`, `num_points` and `rtol`, none of which are in its signature. If clustering near the criticals is wanted, I can look into it as a separate issue, together with that stale docstring.

ViniciusTxc3 (approved):

